PR #1020, opened 2015-01-08
[#1015] Extended usage of project.last_update

kardan:
Extended usage


kardan:
@rumca Once this is merged we can take another go at testing UAT (mention so you get notified),

Issue #895, opened 2014-11-06
Streamline included assets


kardan:
Based on feedback from @iperdomo we should review the urls for included assets. For script tags we can remove the type attribute (https://github.com/akvo/akvo-rsr/pull/883#issuecomment-61636161) & for assets in general we can use protocol-relative URL (http://www.paulirish.com/2010/the-protocol-relative-url/)

Examples of using type with script: 
https://github.com/akvo/akvo-rsr/blob/feature/rsr_v3/akvo/templates/inclusion_tags/maps.html#L2
https://github.com/akvo/akvo-rsr/blob/feature/rsr_v3/akvo/templates/base.html#L21
https://github.com/akvo/akvo-rsr/blob/feature/rsr_v3/akvo/templates/myrsr/my_details.html#L151

We should do a full search in the repo to review if there is more of them.

PR #1027, opened 2015-01-12
[#895] assets

kardan:
Organised included assets


akvo/rsr/views/my_rsr.py
```diff
@@ -29,7 +29,7 @@ def my_details(request):
 
     profile_form = ProfileForm(
         initial={
-            'email': request.user.email,
+            'emailform': request.user.email,
```
KasperBrandt: Why this change?

kardan: Might be a mistake on my part. Hade to debug and later realized it was in the Ajax url. Will review when I can

KasperBrandt: Easy to fix though; just name it back to 'email' and all works again :)..

Issue #1047, opened 2015-01-22
Move IATI codelists to seperate app


KasperBrandt:
- [x] Add new 'codelists' app
- [x] Add models
- [x] Add data
- [x] Replace codelists in RSR
  - [x] Models and admin
  - [x] Templates

PR #1053, opened 2015-01-26
[#1047] Codelists app

KasperBrandt:
- Added new codelists app, with models
  - Also, codelists locally stored in Python files and scripts for generating these
- Data migration for loading in all codelists data
- Updated RSR models and admin with new codelists


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1046, opened 2015-01-22
Sign in always points to 'my details' page


KasperBrandt:
Even when it should be redirecting to another page.

For example, after clicking the 'Add an update' button and not being logged in, the user is redirected to the 'my details' page instead of the 'add update' page.

PR #1054, opened 2015-01-27
[#1046] Small fix for correct sign in redirect

KasperBrandt:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
Testplan:
Go to a project page without being authenticated and hit the "Add update button". After completed authentication the user should end up authenticated on the add update page. (Providing that the user can add update that is).

Issue #1101, opened 2015-02-16
Small code enhancements towards better performance


iperdomo:
* Piwik analytics `<script>` block should be the last thing on the page, to avoid blocking the render

* Use protocol relative when including 3rd party libraries

PR #1102, opened 2015-02-16
[#1101] Small HTML code improvements

iperdomo:
* Move Piwik code to the bottom of the page
* Use protocol relative URL when including 3rd party libs


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1105, opened 2015-02-16
JS code clean up based on JSHint warnings


iperdomo:
Using [JSHint](https://www.npmjs.com/package/jshint) we can detect common errors on JS code.
* Missing semicolons
* Bad expression breaks
* _falsy_ statements
* etc

PR #1110, opened 2015-02-17
[#1105] Fix regression introduced in code cleanup

iperdomo:
* `-` is not allowed in JS identifiers, we need to use array notation to
  access that member


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1116, opened 2015-02-17
Add created_at timestamps to all projects and organisations


zzgvh:
To make RSR KPI generation easier having created_at timestamps for all projects and organisations would help a lot. Since @Lynngre has made a pair of spreadsheets holding this data for almost all objects that don't have it in RSR a simple data migration will do nicely to accomplish this task.

zzgvh:
Bah, bug in the migration. Need to try again.

KasperBrandt:
## Test plan
GIVEN the projects in RSR
WHEN this migration has been done
THEN all projects should have a created_at date filled in

GIVEN the organisations in RSR
WHEN this migration has been done
THEN all organisations should have a created_at date filled in

PR #1117, opened 2015-02-17
[#1116] Populate created at

zzgvh:
Data migration that populates almost all projects and organisations with a created_at timestamp

My running of this on a fresh RSR DB returns about half a dozen projects and one org missing a timestamp from both sources. Those could be hand patched, using the same date as the next lower numbered object. It's not strictly correct, but is probably good enough.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1156, opened 2015-02-20
Location filter bug


KasperBrandt:
The location filter shows different results every time, even though the same link is used. Can be tested by applying a sort after filtering on location.

KasperBrandt:
It has to do with the convert_m49 function in filters.py. Every once in a while it doesn't return the correct results. @kardan you got any clue?

PR #1177, opened 2015-02-23
[#1156] Fixed location filter bug.

kardan:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
Testplan: Make sure sorting on location returns the same results. Can be tested by filtering by Africa and then hit reload a bunch of times. That used to cause the number of projects return to shrink.

PR #1179, opened 2015-02-23
[#1170] Merge RC back into develop

KasperBrandt:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
Migrations fail locally. After review with @KasperBrandt we concluded that we need to update the local vm to match servers. But it will be after 3.0 release.

Issue #1257, opened 2015-02-27
Update text above header


KasperBrandt:
See http://rsr.akvo.org/project/2043/update/8737/

KasperBrandt:
Test plan:

Add a text with markup as update text, it should not show on top of the page. E.g.:
"I believe that several tools will enable effective communication on the programme results, particularly on the water management aspects within the programme. Many of these tools are linked to what the Rain4Food programme (http://www.rain4food.net/) in the Netherlands is doing, where communication on best practices and on rainwater harvesting techniques for food security are encouraged"

PR #1263, opened 2015-02-27
[#1257] Fixed meta og:description elements

kardan:
On project updates pages the ulrize filter was converting URIs to actual
html elements in the meta og:description elements.

Fixed by removing the urlize filter & also limit chars to 500. There is
no hard limit on that field but Google told that Facebook only shows max
300 so this gives us additional space if needed.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1259, opened 2015-02-27
Error on view data from Partner website


adriancollier:
Not sure if this is something within our control or not - can anyone shed any light on this?
![schermafdruk 2015-02-27 10 58 26](https://cloud.githubusercontent.com/assets/1426234/6410486/0d2742b0-be70-11e4-9f12-21194a76980e.png)

KasperBrandt:
I don't think we can do much about it, it's probably something in their PHP script. But the API endpoint they're trying to access (http://rsr.akvo.org/api/v1/project/1605/?format=xml&depth=2) is still working.

KasperBrandt:
Fixed in #1289

KasperBrandt:
## Test plan
No testing needed, the VEI website is up and running again.

PR #1296, opened 2015-03-04
1289 tastypie auth

KasperBrandt:
@kardan Could you check this? It should work now (http://rsr.akvotest.org/api/v1/project/1605/?format=xml&depth=2 now gives a different error when not logged in).

Since this is causing issues with a partner (VEI), whose website is down now (see #1259), we'd like to do a hotfix tomorrow.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1292, opened 2015-03-03
1 or 2 digits behind the decimal point


mato74:
See http://rsr.akvo.org/projects/?title=&location=&status=&organisation=&sector=
Issue: one digit shown where one would expect 2 (.80):

![screenshot 2015-03-03 15 49 57](https://cloud.githubusercontent.com/assets/8478175/6464808/232c9a4c-c1bd-11e4-8da1-8c6986a7ee71.png)

Same on financial info page:
![screenshot 2015-03-03 15 48 37](https://cloud.githubusercontent.com/assets/8478175/6464816/2bf0a9e8-c1bd-11e4-94c3-6884d4e25ac6.png)

adriancollier:
Agreed, we should standardise and always show the decimal points.

At the same time we should align the numbers to the right so that it becomes easier to evaluate the different values displayed.

gabemart:
In addition to the financial info page, should we also align-right the budget on the project listing page?

Current:

![budget-alignleft](https://cloud.githubusercontent.com/assets/2589977/6597121/26ff7ef6-c7f2-11e4-81ce-b5a6a5c95c86.png)

With align-right:

![budget-alignright](https://cloud.githubusercontent.com/assets/2589977/6597123/2df5df7a-c7f2-11e4-8c25-3c8119bbfe74.png)

I vote yes

gabemart:
@KasperBrandt did we decide against adding decimals on the project listing page too, or should I do that now?

KasperBrandt:
@gabemart We haven't really decided on that. I tried it this morning and then realised that there's a limited space for numbers there.. But I think we should, in order to prevent the ".8" cases (see first image in this issue). So yeah, go ahead (same goes for the right alignment).

adriancollier:
I think the safest option for consistency is to always display them.

gabemart:
There remains the question of how to align on narrower viewports - right alignment puts the currency amount very far from their labels. In my opinion makes things harder rather than easier to compare. Shall I only adjust the alignment on larger viewports for now?

![narrowviewport](https://cloud.githubusercontent.com/assets/2589977/6597563/865e5572-c7f5-11e4-939e-86720d60375f.png)

Another approach would be to narrow the width of the entire column of content, like so: 

![financecolumn](https://cloud.githubusercontent.com/assets/2589977/6597731/66d22f2a-c7f6-11e4-9af8-dc08ccfb034a.png)

[...]

PR #1329, opened 2015-03-11
[#1292] Update finance page

KasperBrandt:
@gabemart (or @loicsans) Could you update the alignments on the finance page? The amounts are now all aligned to the left, but should be aligned to the right so that they're easier to compare.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1335, opened 2015-03-13
Feature/1334 docs

adriancollier:
All the docs have been removed from /docs and I have placed a redirection README.md page there to direct people to the right location.

I've also tidied up and inserted all the docs into the wiki so we have not lost any content in this change.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1336, opened 2015-03-16
[1018] Disabled DDT by default

kardan:
Since the Django Debug Toolbar is slow, we should not enable it by
default. Hence commenting out the section in the local settings template.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1337, opened 2015-03-16
1333 gulp errors

kardan:
Besides adding plumber this also updates dependencies. Have to admit that I'm no nom expert so I first did npm update which I think updates my global gulp. Then I killed the scripts/devhelpers/node_modules dir and while in scripts/devhelpers issued npm install to get new modules.

If there is an error in a sass file or a jsx file file and row should be printed now.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
@loicsans maybe you want to verify that this solved your issue?

Issue #1324, opened 2015-03-11
Upgrade to Django 1.7


KasperBrandt:
- [x] Check removed features: https://docs.djangoproject.com/en/1.7/releases/1.7/#features-removed-in-1-7
- [x] Upgrade TastyPie to v0.12.1: http://django-tastypie.readthedocs.org/en/latest/release_notes/v0.12.0.html
- [x] Upgrade from South: https://docs.djangoproject.com/en/1.7/topics/migrations/#upgrading-from-south

adriancollier:
Any estimate on the effort needed here?

KasperBrandt:
Without checking the removed features, this can be done in an hour. However, we have to make sure that there are no other South migrations in the same release, since all migrations have to be removed (they're replaced by Django's internal migrations). So we should move to Django 1.7 at the beginning of a new milestone so that from that point on we can use the new migrations instead. I'll discuss with @kardan if it's possible for the 3.2 release.

kardan:
@KasperBrandt do you see any value in preserving the south table? Otherwise maybe nuke that.

KasperBrandt:
## Test plan
No testing needed.

PR #1368, opened 2015-03-24
[#1324] Django 1.7

KasperBrandt:
@kardan We only need to update the test db so that it contains User.is_support. Can you take a look at that?

Other than that, this should work.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1351, opened 2015-03-18
IATI export v2


KasperBrandt:
See https://github.com/akvo/akvo-product-design/tree/master/RSR/Features/18-IATIExport

- [x] V2.01 Export
- [x] V2.01 Checks
- [x] Fix admin in Django 1.7 for IATI codelists
- [x] UI for creating organisation IATI file
- [x] Add checks to UI
- [ ] < V2 Exports
- [ ] < V2 Checks

KasperBrandt:
## Test plan

GIVEN the MyRSR section
WHEN a user is a superuser OR a rsr admin OR an organisation admin
THEN the My IATI section should be visible

GIVEN the My IATI section
WHEN the user is not allowed to view the My IATI section
THEN a permission denied page should be shown

GIVEN the My IATI section
WHEN the user is a superuser OR rsr admin OR connected to multiple organisations
THEN a select organisation page should be shown

GIVEN the My IATI section
WHEN the user has one approved organisation and is organisation admin
THEN the select organisation page should be skipped and the create IATI file page should be shown directly

GIVEN the create IATI file page
WHEN the selected organisation does not have any reporting projects (projects that have the organisation as reporting organisation)
THEN no projects should be shown

GIVEN the create IATI file page
WHEN the selected organisation has at least one reporting project
THEN these projects should be shown with checkboxes
AND green when all checks are performed successfully
AND red with error explanations when at least one check has failed

GIVEN the create IATI file page
WHEN the user selects at least one project AND creates a IATI file
THEN a IATI file will be created
AND the status or the IATI file itself will be shown on the right side of the page

GIVEN the organisation main page
WHEN a public IATI file is created for the organisation AND the organisation allows public IATI files to be shown on their organisation page
THEN the IATI file should be shown on the page

PR #1382, opened 2015-03-31
[#1351] IATI export v2

KasperBrandt:
First version of IATI export for v2.01 (see issue for test plan).

@gabemart There's a bit of inline styling in ```/templates/myrsr/my_iati.html```, as well as the use of the ```udpateComponent``` class. Could you take a look and fix that properly in the styling files?
Styling in general should be looked at as well btw.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
To me everything seems to work fine. Might be a good thing to later when tests are in mainline add some testing on valid and invalid "raw data" but that is for later.

Issue #1079, opened 2015-02-05
Add user approval to history logs


KasperBrandt:

KasperBrandt:
## Test plan
GIVEN the User Management page
WHEN an employment is approved by a user
THEN this should be shown in the employment admin (user employment admin -> history)

PR #1396, opened 2015-04-01
[#1079] Add admin log entry when approving an employment in MyRSR

KasperBrandt:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1412, opened 2015-04-07
1384 project rest endpoint for up

zzgvh:
Note that the main new functionality, being able to include a specification for a thumbnail in an API GET, is generic, i.e. it works for all models where we have an image: Project, ProjectUpdate and Organisation.

### DIFF
--- a/akvo/rest/fields.py
+++ b/akvo/rest/fields.py
@@ -7,7 +7,6 @@
 
 import base64
 import imghdr
-import os
 import six
 import uuid
 
@@ -16,8 +15,7 @@ from django.utils.encoding import smart_text
 
 from rest_framework import serializers
 from rest_framework.fields import ImageField
-
-from akvo import settings
+from sorl.thumbnail import get_thumbnail
 
 
 class NonNullCharField(serializers.CharField):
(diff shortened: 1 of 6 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
From a quick glance this looks fine, just have a few comments that popped up in my head.
- the "width" & "height" fields are a bit general. Why not image_thumb_height & image_thumb_width as with "image_thumb_name"?
- Do the up app use the same current image in the "list" as on the "project page"? Or maybe the App does a new API call for the "project page" data. Otherwise it might make sense to return an array of [current_image, small, larger].
- Maybe we should have a default cached version & not use the "raw" image. My guess is that an API user might use the current_image if the cached image version needs additional parameters.

stellanl:
Up currently uses the same thumbnail for everything.

zzgvh:
Yep, good comments @kardan. I've updated the code with a new more flexible API to generate thumbs. One question, what size should the default thumb have?

KasperBrandt:
I reckon the same as TastyPie uses at the moment: 160x120.

Issue #1452, opened 2015-04-14
Fix import scripts


KasperBrandt:
Due to the 1.7 upgrade, we need to call django.setup() in every stand-alone script. See http://django.readthedocs.org/en/latest/releases/1.7.html#standalone-scripts.

Issue #1378, opened 2015-03-31
Add IATI import for RVO


KasperBrandt:
If possible, copy the IATI import scripts of RAIN or Cordaid for RVO. The purpose is only to show that we can import their projects on the Test server, and not all information needs to be correctly imported, so that we can show how their projects would look in RSR.

PR #1455, opened 2015-04-14
[#1378 #1452] Add RVO import scripts and fix existing import scripts

KasperBrandt:
No need to check thoroughly. The RVO script will only be used on Test, so that they can see what an eventual Akvo Page for RVO would look like if we would import their IATI data.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1459, opened 2015-04-14
1453 password reset

kardan:
The /sign_in/ resource for password reset cause a 500 error.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

kardan:
Managed to commit with only the commit message "header" so here is what I cut out of the b4c45fe message:
Since we don't have the sites framework in place we need to override domain in the PasswordForm  save method.

Issue #1456, opened 2015-04-14
Paypal donations broken


rumca:
When a user performs the following steps:

* Click to donate to a project
* Select donate with paypal
* Fill out details for donation
* Click next

The following error is thrown:

```
AttributeError at /project/2797/donate/paypal/
'WSGIRequest' object has no attribute 'domain_url'
```

PR #1461, opened 2015-04-14
[#1456] Fixed broken donations.

kardan:
The donation machinery was using a domain_url variable that the old
middleware applied to the request. Now just use the HTTP_ORIGIN instead.


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1501, opened 2015-04-20
[fixes #1498] Fix search text when no search results in MyRSR

### LINKED ISSUE #1498: Really minor issue - search text display when there are no results
This one is really minor, but when a user searches for something on MyRSR and no results are found. Then the text detailing the search still has the `?q=` query string in it, just for the sake of clarity i think we should strip that out:

screen shot 2015-04-15 at 12 26 21

KasperBrandt:


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

PR #1520, opened 2015-04-24
#1519 admin search

### LINKED ISSUE #1519: Search for projects in the admin can get super slow
With the help of ”slow query logs" @orifito found this gem:
*Query*: https://dpaste.de/1mWv
*From*: GET /admin/rsr/project/?q=urban+and+rural+water+supply

So searching in the admin can take up to 339814.442 ms. If someone issue one of these queries that Gunicorn worker will be occupied for 4+ mins. We got 5 of those workers, so 5 searches and RSR is unable to respond on further requests until one query return.

From an explain of the query on UAT we got:
*Explain*: https://gist.github.com/kardan/0ac4df1d5c9ba2623d04

kardan:
Fixes #1519

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ data/*
 !data/README.md
 !data/dump.sh
 !data/load.sh
+!data/simple-load.sh
 
 # Dev environment
 .vagrant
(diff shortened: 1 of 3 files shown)


iperdomo:
## Checklist

* [ ] Test plan
* [ ] Copyright header
* [ ] Code formatting

Issue #1559, opened 2015-05-08
Possible performance improvements?


rumca:
Please close this issue if none of the below can be optimised.

Two areas I've noticed where response times seem a bit slower than the rest of the site, and could perhaps be sped up:
* The project list widget (on the grab a widget page) for organisations with more than a trivial amount of projects takes quite a while to load - it depends on the organisation but Cordaid for example takes well over 60 seconds and 504s
* The project listing page in the admin (`admin/rsr/project/`) takes slightly longer than the other options to load - ~4 seconds for me

kardan:
On the list widget we can try and optimise the query and rendering speeds but since there is no pagination on the widget we will always be at risk with an organisation with a lot of projects. Not really sure on how to deal with that. Possible solutions might be:
- Add pagination
- Use the API to pull in batches
- Limit number of projects in the widget (do you really want 1000 projects in that little window?)

kardan:
After talk with @loicsans concluded that long term solution is pagination but short term we'll cap number of projects to 100 + add a note for when list is capped.

PR #1582, opened 2015-05-26
#1559 admin project directory

kardan:
The odd use case is still slow :-(


kardan:
Testplan:
Hit the project list in the admin and make sure it loads reasonable fast.

PR #1584, opened 2015-05-27
Updated INSTALL information

peeb:
Top level INSTALL information contained 2 dead links. These have been replaced with current documentation links.


peeb:
Maybe it should be a link to `DEVELOPMENT.md` actually - that seems to be the most up-to-date documentation.

Anyway. I was just curious.

kardan:
Docs needs to be updated, good point there.

PR #1585, opened 2015-05-27
Updated setup_etc_hosts.sh devhelper script

peeb:
The script now uses printf in place of echo and adds newlines at the
start end end of the Akvo RSR entry. This clearly separates the Akvo
RSR-specific entry from other entries which may be in the user's
/etc/hosts file.


orifito:
@peeb Good point to separate RSR-specific entry in /etc/hosts. However, we also use that script for some other boxes, and using that approach for all of them wouldn't produce a more clearly hosts file.

peeb:
Ah OK thanks @kardan. I'll stop messing around now :-)

orifito:
@peeb hehe! Actually we are really happy to receive contributions, thanks!

PR #1589, opened 2015-05-29
Minor changes to vagrant README

peeb:
For example. MySQL is no longer used in the RSR stack so references to
it were removed. vagrant command line example updated for latest version
syntax.


peeb:
It might also be an idea to remove the `mysql.sh` script or replace it with a `(postgre)sql.sh` script to avoid confusion for newcomers.

PR #1614, opened 2015-06-09
[Fixes #1508] Project counts are now links

### LINKED ISSUE #1508: Organisation page, overviews
Looking at an organisation page, e.g. http://rsr.akvo.org/organisation/1333/, it is now possible to click on the number of projects or updates that are linked to that organisation. However, it is not obvious that you need to click on the blue text, e.g. 'Projects in RSR'. It would make more sense if you could (also) click on the numbers in order to see the active/need funding/complete/cancelled/total projects.

kardan:
If project buckets are not empty they are now links.


KasperBrandt:
I think we should add the same functionality to the TOTAL UPDATES list. Agree?

Issue #1615, opened 2015-06-11
Small user interface enhancements


KasperBrandt:
See https://github.com/akvo/akvo-product-design/issues/104

gabemart:
- [ ] Change 'Update list' into 'Search'


- [ ] Financial overview page
  - [ ] Remove cents display
  - [ ] Add total amount funded
  - [ ] Add project budget start and end dates


- [ ] Full data report changes
  - [ ] Fix alignment
  - [ ] Add 'Target group'
  - [ ] Rename "Current situation"
  - [ ] Change "other" labels
  - [ ] Export data button


- [ ] Project page - add budget amount and money needed


- [ ] Behaviour of enter key in when writing a project update


- [ ] "Upd..." issue

- [ ] Replace myrsr iati text

gabemart:
Here is my proposal on how to reorganize the finance page. It is slightly different than what you suggested @lauraroverts - I think it might be a bit redundant to include "total funding" in both the 2nd and 3rd column. I propose moving Donations + Current funders to the second column, then showing the total funding + the breakdown of that figure in the 3rd column

![funding mockup](https://cloud.githubusercontent.com/assets/2589977/8212527/15b9b6b0-1514-11e5-92bb-ec4d4484d0d4.png)

@lauraroverts @KasperBrandt thoughts?

KasperBrandt:
@gabemart Looks great!

lauraroverts:
Perfect! Makes much more sense than what I proposed. 
One remark: In the 3th column it now says 'budget' for the total amount of the 'project financial details'. Is that obvious? Shouldn't we than also change 'Project financial details' and 'Budget'  into 'Project budget'?. To use exactly the same words?

gabemart:
@lauraroverts I like that idea, I agree it would be clearer

gabemart:
![pb_mockup](https://cloud.githubusercontent.com/assets/2589977/8226835/2401ff4e-159a-11e5-9b67-e76b1303c384.png)

lauraroverts:
Nice! 
One remark again: In this example it is possible to fund more than the project actually needs (2 dollar). That should not be possible. But I believe there is another issue for that.

gabemart:
Yep - previously when a project was funded over its budget, the "funds needed" showed up as a negative figure (in this example it would have shown as $-2.00). I have changed that to show as $0 needed when a project is over funded.

[...]

PR #1622, opened 2015-06-18
[#1615] Small user interface enhancements

gabemart:
Resolves most of the issues in #1615. 

Exceptions are budget item start and end dates, and showing previous budgets


KasperBrandt:
Oh, I just noticed one more thing; could you add the donation fees as well? Something like this maybe:

ANONYMOUS:           (excl. fee: $10) $120

gabemart:
@Kasper added now. Took some finagling to get it looking right on tablets

Issue #1617, opened 2015-06-11
Translate feature Akvo Pages


lauraroverts:
See https://github.com/akvo/akvo-product-design/issues/106

To clarify:
- Only for Akvo Pages
- Not sure where to put this yet

kardan:
@loicsans we will need UX/UI work on how to deal with two sets of language pickers.

kardan:
Test plan:

With the minimal test dataset & the following entry to /etc/hosts (local host computer)
192.168.50.101 gooddeeds.localakvoapp.org

The Good deeds RSR page should be visible at gooddeeds.localakvoapp.org

If one enable the Google translate bar in the admin and then RSR should display a Google dropdown top right.

KasperBrandt:
There are some styling issues, which @gabemart is looking into.

rumca:
@gabemart this might be the intended styling, but just double checking. With our own translation drop down the header looks like:

![screen shot 2015-06-24 at 10 00 47](https://cloud.githubusercontent.com/assets/3003031/8326424/194ce352-1a59-11e5-89bd-f9cac9ecf95f.png)

However when I enable the google widget the top padding seems to ramp up quite a bit:

![screen shot 2015-06-24 at 10 01 09](https://cloud.githubusercontent.com/assets/3003031/8326442/34900f40-1a59-11e5-82c2-7e4be6d2e211.png)

KasperBrandt:
@rumca Yup, that's intended. When you select a language in the Google Translate widget, it inserts a row above the page that we cannot control. Other than the current solution, we can't do much about it I'm afraid..

PR #1623, opened 2015-06-22
#1617 google translate

kardan:
Test plan:

With the minimal test dataset & the following entry to /etc/hosts (local host computer)
192.168.50.101 gooddeeds.localakvoapp.org

The Good deeds RSR page should be visible at [gooddeeds.localakvoapp.org](http://gooddeeds.localakvoapp.org)

If one enable the Google translate bar in the admin and then RSR should display a Google dropdown top right.


kardan:
@gabemart would you mind having a look at https://github.com/akvo/akvo-rsr/blob/ede0ac9373baf585182a3e2a32bd751a4b1f55b6/akvo/templates/base.html#L5 I'm not sure if there is a more clean way to make the Google translate bar and the RSR navbar happy.

PR #1628, opened 2015-06-23
#1617 google translate

kardan:


KasperBrandt:
Already in here: https://github.com/akvo/akvo-rsr/pull/1626

PR #1634, opened 2015-06-29
[Fixes #1608] more + links display correct counts

### LINKED ISSUE #1608: Organisation double counting
See screenshot: project page says 'reporting partner ICCO + 2 other partners'. If you hover over the (+2 partners) it shows 2 organisations, of which ICCO is also part. It seems that partners with multiple roles are counted double. 

It would make sense to show 'ICCO + 1 other partner'

screen shot 2015-06-04 at 10 29 50

kardan:
Added a 'more_link' template tag to generate the links and made logic
simpler using built in Django template tags.


KasperBrandt:
@kardan Looks good with the templatetag. However, maybe we should not display the link when there's 0 (or -1) partners:

screen shot 2015-07-10 at 09 26 08

Issue #1682, opened 2015-07-23
Photo credit field character length


Lynngre:
The 'photo credit' field currently only allows for 25 characters (although according to @KasperBrandt  it's set to 50).

http://rsr.akvo.org/en/project/projectid/add_update/

This field should allow for more characters not only because it's  even too short of full name right now, bu also often a website may be added as a photo credit. (eg. http://microlancer.lancerassets.com - is the one I was trying to add http://rsr.akvo.org/en/project/3324/update/10642/ ).

## Test plan
GIVEN the add update page
WHEN filling in a new update
THEN it should be possible to enter 75 characters in the 'photo credit' and 'video credit' fields

KasperBrandt:
Fixed in #1701

PR #1705, opened 2015-08-03
[Fixes #1406] Make organisation names unique

### LINKED ISSUE #1406: Make organisation name unique
We need to figure out what to do with the difference between organisation name and long name. However, if we want users to be able to add organisations, we should first make sure that we don't get any double entries of organisations.

## Test plan
GIVEN the organisation admin
WHEN adding a new organisation OR changing an organisation's (long) name
THEN it should not be possible to add an organisation with a name that already exists in the database

GIVEN the organisations in RSR
WHEN looking the organisations directory
THEN all organisation names should be unique
AND all organisation names that weren't unique before, should have an '(2)' or '(3)' attached

Note; since the name sometimes became too long (e.g. organisation.name only allows 25 characters), the organisation names could have been shortened as well. E.g. "Catholic Relief Services" become "Catholic Relief Serv (2)".

KasperBrandt:
[Fixes #1406] Make organisation names unique

Note: has a dependancy on the migration of #1682.

### DIFF
--- a/akvo/rsr/forms.py
+++ b/akvo/rsr/forms.py
@@ -244,7 +244,7 @@ class ProjectUpdateForm(forms.ModelForm):
     photo_credit = forms.CharField(label='', required=False, widget=forms.TextInput(attrs={
         'class': 'input',
         'size': '25',
-        'maxlength': '25',
+        'maxlength': '75',
         'placeholder': _(u'Photo credit'),
         }))
     video = forms.CharField(required=False, widget=forms.TextInput(attrs={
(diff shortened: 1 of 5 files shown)


zzgvh:
I would start with (2) when fixing the names. Thus if three orgs are named **Foobar**, the first keeps its name and the others are named **Foobar (2)** and **Foobar (3)** respectively.

Apart from that the migrations run smoothly.

KasperBrandt:
Yup, that's makes it more clear indeed. Updated the PR now.

Issue #1736, opened 2015-08-17
Locations data migration fails in some cases


zzgvh:
## Test plan
GIVEN the RSR database
WHEN considering all location models (project, project update and organisation)
THEN there should be no objects with the project, project update or organisation set to NULL

## Issue description
Migration 0015_auto_20150714_0853 fails for me when run on my vagrant with data from the live server. Seems there's an unclosed transaction that triggers the failure.

The fix for this is to split the migration in two parts, one that only does the data migration and one that does the schema migration.

KasperBrandt:
Fixed in #1737

PR #1737, opened 2015-08-17
[#1736] Fix data migration for NULL location_targets

zzgvh:
The data migration to remove location objects without location targets
fails when run on the live server dataset. The failure occurs in the
following schema migrations included in the same migration file.

Fix by splitting the migration in two.


KasperBrandt:
@zzgvh I've added a short test plan to #1736

PR #1741, opened 2015-08-18
[Fixes #1740] Send mail when project is published

### LINKED ISSUE #1740: Prevent automatic publishing of projects
## Test plan
GIVEN a project that is going to be published
WHEN the project is published
THEN a notification mail will be sent to Kasper

## Issue description
Somehow there are projects (mainly from Commonsites and Akvo) that get published automatically, even though they shouldn't be. This should be prevented and fixed.

Note; I can't find the reason why this happens. I suspect the API, but then again, this only happens for Akvo and Commonsites projects. Therefore I'll monitor it for now.

KasperBrandt:
Still don't know why some projects get published suddenly, so I'll now receive an email whenever a project is published. Just to check when it happens.


zzgvh:
Leaving the branch. Might be used to remove the code when we figure out the issue.

PR #1742, opened 2015-08-18
[Fixes #1713] Use the meta name='Description' tag

### LINKED ISSUE #1713: Meta description on core RSR project list not correct
## Test plan
Can't test, since our test environments aren't indexed by Google. But all pages on core RSR should have a ```<meta name='Description' ... />``` tag now.

## Issue description
Search for Akvo RSR on Google, and it displays the text of a project, probably the top project at the time of indexing. Should be a general RSR text.

KasperBrandt:
I hope this'll fix the issue.. Can't really test, but as far as I can tell Google doesn't use the ```og:description``` property. I think Facebook does use that though, so I've left it it. But also added the normal ```description``` tag that I believe Google uses.

### DIFF
--- a/akvo/templates/base.html
+++ b/akvo/templates/base.html
@@ -10,28 +10,28 @@
     <meta property="og:site_name" content="Akvo RSR"/>
     <meta property="og:url" content="http://{{request.META.HTTP_HOST}}{{request.path}}"/>
     {% if update %}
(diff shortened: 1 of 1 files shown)


zzgvh:
Added some indents to the meta tag section. Merging anyway.

Issue #1750, opened 2015-08-19
Partnerships with no organisations should not be allowed


KasperBrandt:
## Test plan
GIVEN the project editor
WHEN the project has a funding partnership without an organisation
THEN an error should be displayed that the organisation field isn't filled in

zzgvh:
Actually I think I disagree with the solution here. Is there any scenario where it's valid to have a funding amount without a funding organisation? If not then there should be validation preventing this data from being entered.

PR #1754, opened 2015-08-20
[Fixes #1749] IATI check bug

### LINKED ISSUE #1749: Bug in IATI checks for planned disbursements
## Test plan
GIVEN a project with a planned disbursement
WHEN the IATI checks are performed (MyRSR -> My IATI -> Select organisation -> Perform checks)
THEN the checks should not crash

## Issue description
There is a small bug in the IATI checks for planned disbursements, see http://sentry.support.akvo-ops.org/rsr/test/?url=http://rsr.test.akvo.org/rest/v1/project_iati_check/3522/.

KasperBrandt:
Fixes #1749 

- Change disbursement.amount to disbursement.value

### DIFF
--- a/akvo/iati/checks/v201.py
+++ b/akvo/iati/checks/v201.py
@@ -434,7 +434,7 @@ class V201Checks(object):
         checks = []
 
         for pd in self.project.planned_disbursements.all():
-            if not pd.amount:
+            if not pd.value:
                 checks.append((u'error', u'planned disbursement (id: %s) has '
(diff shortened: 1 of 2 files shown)


zzgvh:
Same as #1750, I don't think there's any point in allowing the saving of a partial Partnership object, i.e. one without a project and an organisation. It just makes for more invalid project states.

I know I haven't been in the discussion around this, so enlighten me if I'm out of line :wink:

KasperBrandt:
@zzgvh Also updated :).

PR #1755, opened 2015-08-20
[Fixes #1750] Fix project funding template

### LINKED ISSUE #1750: Partnerships with no organisations should not be allowed

KasperBrandt:
Fixes #1750


zzgvh:
See comment on #1750

KasperBrandt:
@zzgvh Updated it now, but also with the IATI partner types branch (because of migrations). Anyhow, the Partnership.organisation field is now mandatory.

zzgvh:
I can still save a partnership with only a type:

sceenshot

and I'm not sure there's any point allowing that?

KasperBrandt:
Ah, it doesn't always work indeed :(.. Looking into it now.

PR #1756, opened 2015-08-20
[Fixes #1659] Added IATI activity ID of budgets to finance page

### LINKED ISSUE #1659: Show IATI ID funding streams
## Test plan
GIVEN the project finance page

WHEN a project has multiple funding partners
THEN the __total__ amount should not be stated underneath every funding partner

WHEN a funding partner also has an IATI activity ID specified
THEN that ID should be displayed underneath the name and funding amount

## Issue description
Back in 2013 Aqua for All showed IATI IDs of funding streams from DGIS. These were the following three IATI IDs:
PPP contract :    NL-1-PPR-23062
MA contract:     NL-1-PPR-19884
TMF contract:   NL-1-PPR-12386

Currently these IATI IDs are not shown on the project pages anymore (not in the admin, nor at the project page). Can we retrieve these IATI IDs? It concerns around 140 Aqua for All projects.

KasperBrandt:
Fixes #1659

### DIFF
--- a/akvo/rsr/static/styles-src/main.css
+++ b/akvo/rsr/static/styles-src/main.css
@@ -243,9 +243,15 @@ dl.dl-horizontal dt {
   dl.dl-horizontal dt.totalFinance {
     margin-top: 30px;
     font-weight: bold; }
+  dl.dl-horizontal dt.funders {
+    margin-top: 5px; }
 dl.dl-horizontal dd.totalFinance {
   margin-top: 30px;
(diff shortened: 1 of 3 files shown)


zzgvh:
Yep, the IATI IDs show up as they should. 

screenshot_91

But the layout looks a little wonky, why the duplication of the organisations' name?
screenshot_92

Also it's a bit unclear how an organisation can be a funder without a funding amount...

PR #1786, opened 2015-08-31
[#1785] Cached API results of local storage

### LINKED ISSUE #1785: Local storage full in Chrome
## Test plan
GIVEN the project editor
WHEN the local storage is full (e.g. Chrome)
THEN all typeaheads should still be loaded

## Issue description
When trying to store the API responses in Chrome, this error shows up:

```Uncaught QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of 'cachedAPIResponses' exceeded the quota.```

KasperBrandt:
Fixes #1785

Added a try, except clause to make sure that we don't try to store the API results to the local storage if there is not enough space in the local storage.

### DIFF
--- a/akvo/rsr/static/scripts-src/myrsr-admin.js
+++ b/akvo/rsr/static/scripts-src/myrsr-admin.js
@@ -973,7 +973,12 @@ function loadAsync(url, retryCount, retryLimit, callback) {
 
         lsData = JSON.stringify(localStorageResponses);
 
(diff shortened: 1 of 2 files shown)


kardan:
To test this one would have to fill up local storage (how to and with how much depends on the browser used to test the feature). No docs is needed since it so small and hopefully a corner case.

Issue #1793, opened 2015-09-07
Organisation '+ more' links


KasperBrandt:
## Test plan
Case 1:
When a project only have one partner.
Then no + more link should on the project main page

Case 2:
When a project have one reporting org and another partner.
Then the reporting org + more link with the partner should on the project main page

Case 3:
When a project have one reporting org and the same as partner.
Then the reporting org & no + more link on project main page should be visible

## Issue description
When a project has a reporting organisation and one partner, but a different organisation as a partner, then the '+1 more' link does not display on the project main page.

Probably related to the changes in #1608.

KasperBrandt:
Fixed in #1794

PR #1794, opened 2015-09-07
[#1793] Fix +1 links corner case

kardan:
We exclude Project.reporting_org() org from the partners list to
get counting & display correct. The problem was that a project could
have a reporting org and then a partner that was another org.

### DIFF
--- a/akvo/rsr/static/scripts-src/rsr.js
+++ b/akvo/rsr/static/scripts-src/rsr.js
@@ -48,6 +48,7 @@ $(document).ready(function() {
       "hide": 1000
     }
   });
+  
 
   function getCookie(name) {
     var cookieValue = null;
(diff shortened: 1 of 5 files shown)


kardan:
Testplan:
Case 1:
When a project only have one partner.
Then no + more link should on the project main page

Case 2:
When a project have one reporting org and another partner.
Then the reporting org  + more link with the partner should on the project main page

Case 3:
When a project have one reporting org and the same as partner.
Then the reporting org & no + more link on project main page should be visible

KasperBrandt:
It works on the project main page, but not yet on the project listing page (that wasn't in the issue either..). Can you take a look at that part as well :)?

kardan:
Good point, I'll have a look!

kardan:
@KasperBrandt  Finally managed to get tooltips working in the project directory. Not sure why but I didn't spot the "data-container" & "data-class” attributes on https://github.com/akvo/akvo-rsr/commit/6481d993c1fb586ae4c9f65d43b73f3f0563f7cc#diff-a0683d374328efb31d45e88380829bcdL17 which for some reason blocked the tooltip to appear in the project list but not on the project main page.

PR #1802, opened 2015-09-11
[#1765] remove zepto dependency from project editor

### LINKED ISSUE #1765: Remove Zepto (and jQuery) from project editor
As a first step to completely remove jQuery from RSR, let's remove it completely from the project editor. As far as I know we only use Zepto, but that should also be rewritten to regular Javascript.

gabemart:
Fixes #1765


KasperBrandt:
I think we should leave out the percentage sign in the top progress bar again. Laura specifically asked for that in the beginning. Also write a small test plan, but other than that, looks good :+1: .

PR #1809, opened 2015-09-16
[#1770] IATI import

### LINKED ISSUE #1770: IATI import
## Test plan
1. Take any IATI file from the [IATI registry](http://iatiregistry.org)
2. Make sure the reporting organisation is in RSR, with the correct IATI Organisation ID and 'Reportable' set to True.
3. In the old admin, add a new IATI import. Either fill in an external URL, or a local file. The user that is selected will get an email with a summary of the import. *Note that for larger files (e.g. > 100 projects), you might see a timeout. However, the import should continue to run in the background.*
4. When the import is done, it should show up with status completed, and the mail with import details should have been sent. *Note that files up to 100 projects take about half a minute.*

## Issue description
See product design repository: https://github.com/akvo/akvo-product-design/issues/97

KasperBrandt:
Fixes #1770

Note; Test with small files. I am currently testing with files up to 500 activities, but they take a while..


akvo/iati/imports/iati_import_activity.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def log_changes(self):
+        """
+        Log the changes that have been made to the project in the LogEntry model.
+        """
+        if self.changes:
+            message = u'IATI import, changed: '
+            for change in self.changes:
+                message += u'%s, ' % change
+            message = message[:-2] + u'.'
```
zzgvh: Lines 118 - 121 more pythonic:

```python
            message = u"IATI import, changed: {}.".format(
                u", ".join([change for change in self.changes]))
```

akvo/rsr/models/iati_import_log.py
```diff
@@ -0,0 +1,43 @@
...
+
+class IatiImportLog(models.Model):
+    iati_import = models.ForeignKey(
+        'IatiImport', verbose_name=_(u'iati_import'), related_name='iati_import_logs'
+    )
+    project = models.ForeignKey(
+        'Project', verbose_name=_(u'project'), related_name='iati_project_import_logs',
+        blank=True, null=True
+    )
+    severity = models.IntegerField(_(u'severity'), default=1)
```
zzgvh: I think this could be a PositiveSmallIntegerField with a choices attribute that limits the possible values to the severities.

akvo/rsr/models/iati_import_log.py
```diff
@@ -0,0 +1,43 @@
...
+from django.utils.translation import ugettext_lazy as _
+
+from ..fields import ValidXMLTextField
+
+SEVERITY_CODE = {
+    0: _(u'information'),
+    1: _(u'critical error'),
+    2: _(u'value not saved'),
+    3: _(u'value partly saved'),
+}
```
zzgvh: The severity codes should be a tuple of tuples and used for the choices attribute in IatiImportLog.severity field. [Check out the docs](https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.Field.choices).

Suggested:

```python
class IatiImportLog(models.Model):
    INFORMATIONAL = 0
    CRITICAL_ERROR = 1
    VALUE_NOTSAVED = 2
    VALUE_PARTLY_SAVED = 3
    SEVERITY_CODES = (
        (INFORMATIONAL, _(u'information')),
        (CRITICAL_ERROR, _(u'critical error')),
        (VALUE_NOTSAVED, _(u'value not saved')),
        (VALUE_PARTLY_SAVED, _(u'value partly saved')),
    )
    ...
```
This way it's immediately obvious in the code that uses the severty codes what a particular severity represents, rather than having a plain integer that you have to trace back to it's origin. Also by putting these definitions inside the class, it becomes very easy to import it all since it's part of the package when you ```from iati_import_log import IatiImportLog```!

akvo/rsr/models/iati_import_log.py
```diff
@@ -0,0 +1,43 @@
...
+    class Meta:
+        app_label = 'rsr'
+        verbose_name = _(u'IATI import log')
+        verbose_name_plural = _(u'IATI import logs')
+
+    def show_severity(self):
+        if not self.severity in SEVERITY_CODE.keys():
+            return _(u'unknown severity')
+        else:
+            return SEVERITY_CODE[int(self.severity)]
```
zzgvh: show_severity() now doesn't need the check the value of self.severity, we **know** it can only be one of the four SEVERITY_CODES choices.

To return the label given the severity code we can do this:

```python
def show_severity(self):
    return dict(map(lambda x: x, SEVERITY_CODES))[self.severity]
```

akvo/rsr/models/iati_import.py
```diff
@@ -0,0 +1,62 @@
+# -*- coding: utf-8 -*-
```
zzgvh: ```STATUS_CODE``` and ```IatiImport.status``` should get the same treatment as suggested in iati_import_log.py

PR #1836, opened 2015-10-04
[#1515] Filters on listing pages

### LINKED ISSUE #1515: Akvo Pages information (materialized views)
## Test plan
GIVEN core RSR and Akvo Pages
WHEN opening the filters on the project and organisation listing pages
THEN only countries that have results should be shown

GIVEN Akvo Pages only
WHEN opening the filters on the project, organisation and update listing pages
THEN only organisations that are relative to the page should appear in the filter list

## Issue description
At Akvo Pages the dropdowns to filter for an Organisation or Country include all organisations and countries that have got a project in RSR. E.g. http://washalliance.akvoapp.org/projects/. It would be great if at an Akvo Page only those organisations and countries that are linked to the partner's Akvo Page will appear in the dropdown. 

For the Dutch WASH Alliance that would mean only 8 countries show up in the dropdown, because they only work in 8 countries. 

It used to be like this in the old version of RSR.

kardan:
Fixes #1515

Could be extended to use MultipleChoice filters on locations, but UI get's a bit tight.

### DIFF
--- a/akvo/rsr/filters.py
+++ b/akvo/rsr/filters.py
@@ -169,7 +169,7 @@ class ProjectUpdateFilter(django_filters.FilterSet):
     partner = django_filters.ChoiceFilter(
         choices=get_orgs(),
         label=_(u'organisation'),
-        name='project__partners__id')
+        name='user__organisations__id')
 
     sector = django_filters.ChoiceFilter(
         initial=_('All'),
--- a/akvo/rsr/filters.py
+++ b/akvo/rsr/filters.py
@@ -10,14 +10,12 @@ see < http://www.gnu.org/licenses/agpl.html >.
 import django_filters
 
 from copy import deepcopy
-
-from .models import Project, Organisation, Category, ProjectUpdate
-from .m49 import M49_CODES, M49_HIERARCHY
-
+from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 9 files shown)


kardan:
Testplan:

For normal RSR the only notable difference should be that on the list pages only locations that have data should be in the drop down.

On RSR pages:
- Only locations where there is projects & organisations should appear (for updates this is broken since there is no country data but only lat, long - needs to be fixed with reverse geolookup). 
- Only organisations that are relative to the page should appear in the filter lists for projects, updates & orgs.

KasperBrandt:
@kardan I get some errors:

1. On core RSR when trying to filter on organisation on the project listing page, it says ```need more than 0 values to unpack```
2. On core RSR the loading time for the update listing page is very long
3. On core RSR and Akvo Pages when trying to filter on location on the organisation listing page, it says ```int() argument must be a string or a number, not 'list'``` (guess it shouldn't be a multiple choice option there)
4. On an Akvo Page when trying to filter on organisation on the project listing page, and selecting "All" + an organisation, it says: ```invalid literal for int() with base 10: ''```

kardan:
hmm this is odd since I used my dataset from prod to test it. I'll have to review again.

PR #1840, opened 2015-10-05
[#1288] Add map explanation

### LINKED ISSUE #1288: Explanation to maps: this map shows the projects listed on this page.
## Test plan
GIVEN the project, organisation, update listing pages on core RSR
WHEN viewing the map
THEN a message should be shown on the bottom of the map that not all objects are shown

GIVEN an Akvo Page
WHEN the 'All points on maps' setting is set to True
THEN the message should not be shown

## Issue description
There are mixed opinions on the use-case for what pins to display on the maps within RSR - particularly the project map, but this also goes for updates and organisations.

At the moment the map only displays the points that are being shown within the results in the list - so the first page of results. Some users are requesting to be able to see all these elements on the map.

One challenge with this is that this is costly from a system resources point of view.

One potential solution is to provide an option on the listing page(s) to allow the user to choose to add all the results to the map rather than just the first page.

Would need to ensure the wording on this option is clear.

KasperBrandt:
Fixes #1288

### DIFF
--- a/akvo/rsr/static/styles-src/main.css
+++ b/akvo/rsr/static/styles-src/main.css
@@ -773,6 +773,23 @@ h4.detailedInfo {
     #map {
       display: none; } }
 
+#map-warning {
+  margin-top: -30px;
+  z-index: 99;
+  opacity: 0.8;
+  background: white;
+  margin-right: 15%;
+  margin-left: 15%;
+  padding: 5px;
+  border: solid 1px;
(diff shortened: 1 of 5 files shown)


KasperBrandt:
@gabemart Linked this one to you, no rush though.. Implemented it like this now:

zzgvh:
Code reviewed.

PR #1842, opened 2015-10-06
[#795] Fix loop when organisation is content_owner of itself

### LINKED ISSUE #795: Editing permissions of organisations other than your own.
## Test plan
GIVEN the MyRSR user management page (/myrsr/user_management)
WHEN logged in as an organisation admin or user manager
THEN you should be able to see and approve/decline all users from your organisation __AND__ the organisations that have your organisation as a content_owner __AND__ organisations being content owned by that organisation, etc.

GIVEN the old organisation admin (/admin/rsr/organisation/)
WHEN logged in as an organisation admin or project editor
THEN you should be able to see and edit your organisation __AND__ all organisations that have your organisation as a content_owner __AND__ organisations being content owned by that organisation, etc.

## Issue description
This issue is a result of talking to Uddy who's implementing the PIND data import into RSR using the API. I think we need to extend the organisations users belonging to the Partner Admin group is allowed to access to those organisations that have the **content_owner** field set to the organisation of the user.

KasperBrandt:
Fixes #795

### DIFF
--- a/akvo/rsr/models/organisation.py
+++ b/akvo/rsr/models/organisation.py
@@ -315,7 +315,9 @@ class Organisation(TimestampsMixin, models.Model):
         org_set = set()
         org_set.add(self)
 
-        self_content_owned_list = list(Organisation.objects.filter(content_owner=self))
+        self_content_owned_list = list(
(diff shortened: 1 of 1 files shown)


KasperBrandt:
Better solution in #1844

Issue #1772, opened 2015-08-26
Report on indicators 


lauraroverts:
## Test plan
TODO: Extend test plan

GIVEN the 'Results & indicators' tab on the project main page
WHEN the user clicks the 'Add a new update' button
THEN it should be possible to select an indicator period
AND update the value of the indicator period

GIVEN the 'Results & indicators' tab on the project main page
AND the user is an organisation Admin
WHEN the user clicks the 'Edit update' button
THEN it should be possible to edit the update

## Issue description
As part of adding a 'results framework' to RSR. See https://github.com/akvo/akvo-product-design/issues/101

The main focus of this issue is to be able to edit the actual values of an indicator period from within RSR.

PR #1848, opened 2015-10-08
[#1772] Add indicator fields to ProjectUpdate model

KasperBrandt:
After merging, run ```sh manage.sh migrate rsr``` in the devhelpers folder.


gabemart:
Code checked :+1:

PR #1882, opened 2015-10-27
[#1877] Calculate sliders from start of period

### LINKED ISSUE #1877: Calculate updates from baseline, not from actual
## Test plan
There shouldn't be a difference from the original results framework. No specific testing needed.

## Issue description
Currently, the place of the updates on the slider is calculated by going back from the actual. However, this sometimes results in strange behaviour (e.g. the slider being before the baseline). We should calculate starting from the baseline and add up all the previous updates.

KasperBrandt:
Fixes #1877

### DIFF
--- a/akvo/rsr/static/scripts-src/project-main.js
+++ b/akvo/rsr/static/scripts-src/project-main.js
@@ -1223,61 +1223,52 @@ if (firstAccordionChild !== null) {
           return a.id > b.id;
       });
 
-      var actualText, highestValue, periodNode, periodValue, periodTarget, periodBaseline;
-
-      highestValue = 0;
+      var actualText, periodNode, periodValue, periodTarget, periodBaseline, progress;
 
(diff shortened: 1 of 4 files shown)


akvo/rsr/static/scripts-src/project-main.js
```diff
@@ -1223,61 +1223,52 @@ if (firstAccordionChild !== null) {
...
-          */
-          if (periodValue > highestValue) {
-            highestValue = periodValue;
-          }
-          var updateMarker = document.createElement('div');
-          var textSpan = document.createElement('div');
+        if (entry.id <= update.id || update.id === 'add') {
+          /* This update is older (or the same) as the update we're building the dialog for.
+          We need to make a marker for it. */
+          progress += parseInt(entry.period_update);
```
iperdomo: We should use `parseInt(entry.period_update, 10)`

gabemart:
code checked

PR #1883, opened 2015-10-27
[#1881] Fix aggregation status and ascending in project editor

### LINKED ISSUE #1881: Aggregation status and Ascending can't be "No"
## Test plan
GIVEN the Aggregation status (Result) and Ascending (Indicator) in section 5 of the project editor (click 'Advanced IATI fields' to see those)

1. Setting the value to "No" should not give an error
2. Setting the value to either nothing, "Yes" or "No" should save correctly (refresh the page to check if the values actually have been stored correctly).

## Issue description
Setting the aggregation status or ascending to "No" in the project editor results in an error.

KasperBrandt:
Fixes #1881


gabemart:
Code checked :+1:

PR #1904, opened 2015-11-11
[#1900] Slight change in primary project partner

### LINKED ISSUE #1900: Project does not always have a primary partner
## Test plan
GIVEN the project listing page and project main page
WHEN a project has no partners
THEN no internal server error should be shown, but just no partners

GIVEN the project main page
THEN no 'Financial info' should be shown, since we now have a 'Finance' tab

GIVEN the finance tab on the project main page
WHEN an organisation without a logo is a partner
THEN the 'No logo specified' image should not be bigger than 80x60 pixels

## Description
There are situations in which the project does not have a primary partner. Even though this should be a scenario that doesn't occur very often, we do get errors from it now in Sentry: http://sentry.support.akvo-ops.org/rsr/live/group/711/, http://sentry.support.akvo-ops.org/rsr/live/group/713/.

KasperBrandt:
Fixes #1900: 
- Added the first 'can_create_projects' partner as the primary partner. I guess this will hold in almost all cases.

In addition, did two small changes as well..
- Removed the financial info part from the project main page (as there's a tab for that now).
- Set the geometry for the logo of partners without a logo correct on the finance page.

### DIFF
--- a/akvo/rsr/models/project.py
+++ b/akvo/rsr/models/project.py
@@ -512,6 +512,14 @@ class Project(TimestampsMixin, models.Model):
         """
         return self.reporting_partner.organisation if self.reporting_partner else None
 
+    @property
+    def publishing_orgs(self):
+        """
+        Returns the organisations that have the right to publish the project. In other words, that
+        have Organisation.can_create_project set to True.
+        """
+        return self.partners.filter(can_create_projects=True)
+
     def set_reporting_org(self, organisation):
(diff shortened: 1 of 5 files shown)


zzgvh:
Test plan?

KasperBrandt:
Whoops, forgot.. It's here now: https://github.com/akvo/akvo-rsr/issues/1900#issue-116277617

PR #1905, opened 2015-11-11
[#1903] Fix validation of iati_org_id

### LINKED ISSUE #1903: Error in validation when saving an Organisation
When you add or change an IATI organisation ID in the RSR admin, an AttributeError occurs. The reason for this is a bug in the check for duplicate IATI IDs.

zzgvh:
Fixes #1903

Fix an AttributeError when adding or changing Organisation.iati_org_id.


KasperBrandt:
Tested and works :+1:

PR #1906, opened 2015-11-11
[#1901] Fix full data report when partnership has no role

### LINKED ISSUE #1901: Partnership doesn't always have a role
## Test plan
GIVEN the project main page
WHEN the project has a partnership without a role
THEN no error should be shown, but the __ROLE__ of the partnership should be empty

## Description
When a partnership has no role, the report tab of the project main page makes the page crash.

Although this scenario doesn't occur often, it does occur: http://sentry.support.akvo-ops.org/rsr/live/group/709/, http://sentry.support.akvo-ops.org/rsr/live/group/679/.

KasperBrandt:
Fixes #1901

### DIFF
--- a/akvo/rsr/models/partnership.py
+++ b/akvo/rsr/models/partnership.py
@@ -135,7 +135,10 @@ class Partnership(models.Model):
     )
 
     def iati_organisation_role_label(self):
-        return dict(self.IATI_ROLES)[self.iati_organisation_role]
+        if self.iati_organisation_role:
+            return dict(self.IATI_ROLES)[self.iati_organisation_role]
+        else:
+            return None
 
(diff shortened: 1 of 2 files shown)


akvo/rsr/models/partnership.py
```diff
@@ -135,7 +135,10 @@ class Partnership(models.Model):
     )
 
     def iati_organisation_role_label(self):
-        return dict(self.IATI_ROLES)[self.iati_organisation_role]
```
zzgvh: I would use dict.get instead:
```python
def iati_organisation_role_label(self):
    return dict(self.IATI_ROLES).get(self.iati_organisation_role)
```

PR #1907, opened 2015-11-11
[#1902] Performance improvements

### LINKED ISSUE #1902: Degraded performance of project listing page
## Test plan
The project listing page should load fast now, the number of DB queries is reduced greatly. 

Also all projects with a partner should have a primary_organisation (which is visible on the project listing page and project main page). And when a project partner is added / edited / deleted, this organisation should be updated with the following rules:

- If an organisation can create projects, then this becomes the primary organisation
- If an organisation otherwise is reporting organisation, this becomes the primary
- If an organisation otherwise is accountable organisation, this becomes primary
- Else it's just the first organisation

## Description
The project listing page is slower than it used to be. We should check out the cause of this.

KasperBrandt:
Fixes #1902.

Includes the changes of the #1900 branch, that one should be included first.


akvo/rsr/templatetags/rsr_tags.py
```diff
@@ -17,21 +17,35 @@
...
+
+    for partnership in project.partnerships.all():
+        organisation = partnership.organisation
+
+        if partnership.iati_organisation_role_label():
+            label = partnership.iati_organisation_role_label()
+        else:
+            label = _(u'No partner role')
+
+        if not organisation == project.primary_organisation:
```
zzgvh:
```python
if organisation != project.primary_organisation:
```

zzgvh:
The [coll parameter](https://github.com/akvo/akvo-rsr/blob/%231902-performance/akvo/rsr/views/utils.py#L11) here is confusing. It implies some generic collection, but in all cases it's a project query set. I think it should be named projects or project_qs or the like.

PR #1918, opened 2015-11-17
[#1820] Private projects

### LINKED ISSUE #1820: Make (part of) projects private
## Test plan
GIVEN a project that has rsr_project.is_private = True
_Note: this can currently only be set in the 'old' admin or directly in the DB_

WHEN visiting the RSR projects listing page
THEN the project should not appear

WHEN visiting the RSR project updates listing page
THEN the updates of the project should not appear

WHEN visiting the main project page
AND the user has sufficient rights to see the project
THEN the project should be shown with a statement that it is a private project

WHEN visiting the main project page
AND the user does not have sufficient rights to see the project or is not logged in
THEN a 'Permission denied' page should be shown if the user 

WHEN visiting the My projects page in MyRSR
AND the user has sufficient rights to see the project
THEN the project should be shown with a lock in front of it

WHEN visiting the My IATI page in MyRSR
THEN the project should be shown in the list, but with a _(private project)_ indication and unchecked

WHEN looking up the project or any of it's sub-objects in the API
THEN it should state that the project or object has not been found

## To do
- Project editor should include a 'Private' flag. This will be taken along in the RSR / IATI editor split (#1895).
- IATI import should set project to private when this is indicated in a IATI file. This will be taken along in the IATI import improvements (#1828).

## Issue description
Once IATI / RSR / RSR Impact is in place, we need to investigate if it's possible for organisations to make (part of) their projects private. 

This was requested by ICCO --> Since it is expected that in the future more organisations will publish their projects according to the IATI standard. Therfore it is likely that some of their projects need to be private. E.g. because they can't publish data about a certain region due to political or security reasons. If we allow partners to keep part of their projects private, it is more likely they will use RSR for all of their projects.

KasperBrandt:
Fixes #1820

### DIFF
--- a/akvo/rest/views/project.py
+++ b/akvo/rest/views/project.py
@@ -46,12 +46,13 @@ class ProjectViewSet(BaseRSRViewSet):
     __keywords__ (exact, in), and
     __publishingstatus\__status__.
     """
-    queryset = Project.objects.select_related(
-        'publishingstatus'
-        ).prefetch_related(
-            'categories',
-            'keywords',
-            'partners')
+    queryset = Project.objects.public().select_related(
+        'categories',
+        'keywords',
+        'partners',
+    ).prefetch_related(
+        'publishingstatus',
+    )
 
     serializer_class = ProjectSerializer
     filter_fields = {
@@ -107,9 +108,11 @@ class ProjectExtraViewSet(ProjectViewSet):
     __publishingstatus\__status__ (filter on publishing status)
     """
 
-    queryset = Project.objects.select_related(
-        'publishing_status').prefetch_related(
-            'sectors', 'partnerships')
+    queryset = Project.objects.public().prefetch_related(
+        'publishingstatus',
+        'sectors',
+        'partnerships',
+    )
     serializer_class = ProjectExtraSerializer
     paginate_by_param = 'limit'
     filter_fields = ('partnerships__organisation', 'publishingstatus__status')
@@ -126,14 +129,15 @@ class ProjectUpViewSet(ProjectViewSet):
     __publishingstatus\__status__ (filter on publishing status)
     """
 
-    queryset = Project.objects.select_related(
+    queryset = Project.objects.public().select_related(
         'primary_location',
+        'categories',
+        'keywords',
+        'partners',
+    ).prefetch_related(
+        'publishingstatus',
         'updates',
-        'publishingstatus'
-        ).prefetch_related(
-            'categories',
-            'keywords',
-            'partners')
+    )
     serializer_class = ProjectUpSerializer
     paginate_by_param = 'limit'
     max_paginate_by = 100
--- a/akvo/rest/views/project.py
+++ b/akvo/rest/views/project.py
@@ -6,7 +6,7 @@ For additional details on the GNU license please see < http://www.gnu.org/licens
 """
 
 from akvo.rest.serializers.project import ProjectUpSerializer
-from akvo.rsr.models import Project, PublishingStatus
+from akvo.rsr.models import Project
 from ..serializers import ProjectSerializer, ProjectExtraSerializer
 from ..viewsets import BaseRSRViewSet
 
@@ -46,7 +46,7 @@ class ProjectViewSet(BaseRSRViewSet):
     __keywords__ (exact, in), and
     __publishingstatus\__status__.
     """
-    queryset = Project.objects.public().select_related(
+    queryset = Project.objects.select_related(
         'categories',
         'keywords',
         'partners',
--- a/akvo/rest/views/project_update.py
+++ b/akvo/rest/views/project_update.py
@@ -8,14 +8,14 @@ For additional details on the GNU license please see < http://www.gnu.org/licens
 from akvo.rsr.models import ProjectUpdate
 
 from ..serializers import ProjectUpdateSerializer, ProjectUpdateExtraSerializer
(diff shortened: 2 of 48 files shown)


akvo/rest/views/project.py
```diff
@@ -46,12 +46,13 @@ class ProjectViewSet(BaseRSRViewSet):
...
-        ).prefetch_related(
-            'categories',
-            'keywords',
-            'partners')
+    queryset = Project.objects.public().select_related(
+        'categories',
+        'keywords',
+        'partners',
+    ).prefetch_related(
+        'publishingstatus',
```
zzgvh: Why are you prefetching publishingstatus?

akvo/rest/views/project_update.py
```diff
@@ -40,30 +38,29 @@ def get_queryset(self):
...
         project__partners = self.request.QUERY_PARAMS.get('project__partners', None)
         if project__partners:
-            queryset = queryset.filter(project__partners=project__partners)
+            self.queryset = self.queryset.filter(project__partners=project__partners)
         user__organisations = self.request.QUERY_PARAMS.get('user__organisations', None)
         if user__organisations:
-            queryset = queryset.filter(user__organisations=user__organisations)
-        return queryset
+            self.queryset = self.queryset.filter(user__organisations=user__organisations)
+        return super(ProjectUpdateViewSet, self).get_queryset()
```
zzgvh: I think a few lines can be removed here (NOTE: untested code):

```python
    def get_queryset(self):
        """
        Allow simple filtering on selected fields.
        We don't use the default filter_fields, because Up filters on
        datetime for last_modified_at, and they only support a date, not datetime.
        """
        query_keys = [
            'created_at__gt',
            'created_at__lt',
            'last_modified_at__gt',
            'last_modified_at__lt',
            'project__partners',
            'user__organisations',
        ]

        for key in query_keys:
            value = self.request.QUERY_PARAMS.get(key, None)
            if value:
                self.queryset = self.queryset.filter(**{key: value})

        return super(ProjectUpdateViewSet, self).get_queryset()
```

akvo/rsr/views/project.py
```diff
@@ -152,6 +152,18 @@ def directory(request):
...
+    Checks if the user can view a project, otherwise raises a PermissionDenied exception.
+
+    A user can view any public project, but when a project is private or not published the user
+    should be logged in and able to make changes to the project (e.g. be an admin of the project).
+    """
+    if (project.is_private or not project.is_published()) and \
+            user.is_anonymous() or \
+            (not user.is_anonymous() and not user.has_perm('rsr.change_project', project)):
+        raise PermissionDenied
+
```
zzgvh: I'm guessing this could be shortened thus:

```python
    if ((project.is_private or not project.is_published()) 
            and not user.has_perm('rsr.change_project', project)):
        raise PermissionDenied
```
since anonymous users never have rsr.change_project perm.

akvo/rest/viewsets.py
```diff
@@ -22,3 +22,13 @@ class BaseRSRViewSet(viewsets.ModelViewSet):
     permission_classes = (DjangoObjectPermissions, )
     filter_backends = (filters.DjangoFilterBackend, filters.OrderingFilter, )
     ordering_fields = '__all__'
+
```
zzgvh: I think you can eliminate the need for the PublicRSRViewSet entirely by adding the BaseRSRViewSet.get_queryset instead:

```python
    def get_queryset(self):
        """Filter out any private projects."""
        for related_obj in self.queryset.model._meta.get_all_related_objects():
            if related_obj.model == Project:
                self.queryset = self.queryset.filter(project__is_private=False)
                break
        return super(BaseRSRViewSet, self).get_queryset()
```

NOTE: this is untested code!

akvo/rsr/views/project.py
zzgvh: Also we should start trying to use django-rules predicates to test for permissions. _check_project_viewing_permissions() looks like a candidate for this even tho I'd need to read up on django-rules to say exactly how.

akvo/rsr/models/project.py
```diff
@@ -192,6 +192,11 @@ class Project(TimestampsMixin, models.Model):
                                             help_text=_(u'Determines whether this project is an '
                                                         u'RSR Impact project.'))
 
+    # Private projects
+    is_private = models.BooleanField(_(u'is private project'), default=False,
```
zzgvh: I would name the field **is_public**. Then the logic for it becomes similar to is_published.

akvo/rest/views/project.py
KasperBrandt: Since we fetch the publishingstatus as well (in ProjectSerializer):
```publishing_status = serializers.Field(source='publishingstatus.status')```

akvo/rsr/views/project.py
KasperBrandt: I've shortened it now, but haven't looked at the django-rules predicates yet.. We'll have to do a big refactor of the permissions system somewhere next year anyway, to allow project specific permissions.

akvo/rest/viewsets.py
KasperBrandt: Yup, correct. Changed it, although instead of the ```break``` I just return the queryset.

akvo/rsr/models/project.py
KasperBrandt: Yes, true.. Changed now.

PR #1928, opened 2015-11-23
[#1923] Widgets fix for projects without reporting org

### LINKED ISSUE #1923: Project list widget throws an internal server error
## Test plan
GIVEN the 'Grab a widget' page

WHEN a project has no partners
THEN the project list and project map options should not be available

WHEN a project has partners
THEN all widgets should be available and not display an error

## Issue description
The project list widget gives an internal server error. And the map widget as well: http://sentry.support.akvo-ops.org/rsr/live/group/727/. This is due to the fact that they don't have a reporting organisation and the map and project listing widgets expect an organisation as an input.

The reporting organisation here should be changed to the project's primary organisation, and we should create a 404 exception when the project has no partners at all, just in case.

KasperBrandt:
Fixes #1923

### DIFF
--- a/akvo/templates/project_widgets2.html
+++ b/akvo/templates/project_widgets2.html
@@ -29,28 +29,30 @@
                             <input type="radio" name="widget-project" value="project" checked>{% blocktrans with title=project.title %}Use the {{title}} project{% endblocktrans %}
                         </label>
                     </div>
-                    <div class="radio">
-                        <label>
-                            <input type="radio" name="widget-project" value="random">{% trans "Display a random project from an organisation" %}:
-                        </label>
-                    </div>
-                    <div class="form-group">
-                        <select class="form-control" id="widget-organisation" disabled>
-                        <option value="{{project.reporting_org.pk}}">{{project.reporting_org.name}}</option>
-                        {% for partner in project.all_partners.distinct %}
-                            {% if not partner == project.reporting_org %}
-                                <option value="{{partner.pk}}">{{partner.name}}</option>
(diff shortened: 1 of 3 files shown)


akvo/templates/project_widgets2.html
```diff
@@ -29,28 +29,30 @@
...
-                        <select class="form-control" id="widget-organisation" disabled>
-                        <option value="{{project.reporting_org.pk}}">{{project.reporting_org.name}}</option>
-                        {% for partner in project.all_partners.distinct %}
-                            {% if not partner == project.reporting_org %}
-                                <option value="{{partner.pk}}">{{partner.name}}</option>
-                            {% endif %}
-                        {% endfor %}
-                      </select>
-                    </div>
+                    {% if project.all_partners.distinct.count > 1 %}
```
zzgvh: This should be ```{% if project.all_partners.distinct.count >= 1 %}``` right?

KasperBrandt: Yeah, doesn't really matter a lot I think.

The primary organisation is selected by default. In the current implementation you only see the other option when there are actually other organisations to choose from. We could also show it when there's no other organisation besides the primary organisation, although there's not a lot of added value for that :).

zzgvh: But this means if there's only one org associated with the project, the choice of a random project goes away.

KasperBrandt: Yeah, I missed that.. There is indeed a difference in functionality there. Adjusted now.

PR #1958, opened 2015-12-23
[#1005] Invite users

### LINKED ISSUE #1005: Invite User Functionality
## Test plan
GIVEN the user management page
WHEN the user is a superuser, admin, org admin, or org user manager
THEN a new button should appear in the top right corner

GIVEN the new 'invite' button
WHEN clicked
THEN a modal should appear in which it is possible to add and delete rows (invited users) with an email address, organisation, and group

GIVEN that a user or multiple users have been fully filled in
WHEN the invite button is clicked
THEN the user should receive an email that he or she is invited to use RSR
OR the user should receive a notification that his new employment is approved (in case of an existing user)

GIVEN that a user or multiple users have not been fully filled in
WHEN the invite button is clicked
THEN the missing fields should get a red border

GIVEN that the user is a superuser or RSR admin
WHEN the user wants to invite a new user
THEN all organisations and roles can be chosen

GIVEN that the user is not a superuser or RSR admin
WHEN the user wants to invite a new user
THEN only the 'Users' and 'Project editors' roles can be chosen
AND the organisations that they're connected with
AND the organisations that these organisations 'content own'
AND, in case of a paying partner, the organisations that are implementing partners on their projects

GIVEN that the invite has been sent to a user
THEN the fields in the modal should become green
AND after 3 seconds the row should disappear
AND if it's the last row, the modal should disappear

GIVEN that a user receives a new invite
WHEN the activation link in the mail is clicked
THEN the user should be directed to an activation page where it is possible to fill in first name, last name and a password

GIVEN that the activation link is expired (after 7 days)
THEN the user should see an appropriate message _(not sure how to test this easily)_

GIVEN that the activation link is incorrect
THEN the user should see an appropriate message

GIVEN that the invited user has activated his or her account
THEN the user that invited the new user should get a confirmation mail that the invitation was accepted

## Issue description
We need the ability for Organisation Admins or User managers to add users based on their email address, or for superusers and RSR admin to add a user to any organisation.

- [x] Option for Organisation Admins or User managers in MyRSR to invite users to their organisation
- [x] Option to enter one or more email addresses, and set the organisation (organisation + implementing partners for paying partners) and the permissions (User or Project editor) for every email address
- [x] If email address, send an email inviting them to join RSR and complete registration (e.g. fill in first name, last name, password) after email confirmation
- [x] If known email address, send a message to user notifying them that they have been added to an organisation
- [x] Notify the inviter when the invitee has activated his/her account

KasperBrandt:
Fixes #1005


akvo/rest/views/user_management.py
```diff
@@ -0,0 +1,121 @@
...
+    if missing_data:
+        return Response({'missing_data': missing_data}, status=status.HTTP_400_BAD_REQUEST)
+
+    # Check if the user already exists, based on the email address
+    try:
+        invited_user = get_user_model().objects.get(email=email)
+    except get_user_model().DoesNotExist:
+        try:
+            invited_user = get_user_model().objects.create_user(email, email=email)
+            invited_user.set_is_active(False)
```
zzgvh: This is somewhat insecure. It'd be better if ```create_user()``` returned an inactive user that was then explicitly set to active. Otherwise there is a small chance of a user being created as active and then something goes wrong before ```invited_user.set_is_active(False)``` is called.

akvo/rsr/models/organisation.py
```diff
@@ -364,6 +376,13 @@ def support_partners(self):
         in common with, excluding self"""
         return self.published_projects().support_partners().exclude(id__exact=self.id)
 
+    def field_partners(self):
+        """
+        Returns an Organisation queryset of field partners of which self has at least
+        one project in common with.
+        """
+        return self.all_projects().field_partners().distinct()
```
zzgvh: ```Organisation.support_partners()``` uses ```self.published_projects()```. It seems logical to use the same in both methods, although I can't say if ```published_projects()``` or ```all_projects()``` is the correct one.

akvo/urls.py
```diff
@@ -89,6 +89,9 @@
     url(r'^activate/(?P<activation_key>\w+)/$',
         'akvo.rsr.views.account.activate', name='activate'),
 
+    url(r'^activate_invite/(?P<inviting_pk>\d+)/(?P<pk>\d+)/(?P<token_date>[0-9A-Za-z_\-]+)/(?P<token>[0-9A-Za-z_\-]+)/$',
+        'akvo.rsr.views.account.invite_activate', name='invite_activate'),
```
zzgvh: Why not use the same URL and view name?

akvo/templates/registration/invited_user_notification_message.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
zzgvh: Those templates should named inviting_user_notification_*

akvo/templates/registration/invite_user_message.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
```
zzgvh: Should be named invited_user_*

akvo/rest/views/user_management.py
```diff
@@ -0,0 +1,121 @@
...
+    employment, empl_created = Employment.objects.get_or_create(
+        user=invited_user, organisation=organisation
+    )
+
+    # For existing employments, we do not change the group
+    if empl_created:
+        employment.group = group
+    # Approve the employment
+    if not employment.is_approved:
+        employment.approve(user)
```
zzgvh: Approval shouldn't happen here I think, but rather when the user accepts the invite. As it is now we trigger pre- and post-save signals for employment that send emails to admins and the user, possibly creating confusion since we at this stage don't even know the email address works!

akvo/rest/views/user_management.py
KasperBrandt: Yeah, I agree. I've now changed the default setting of the ```create_user()``` function to create inactive users, since it's kind of weird to create active users. I checked and user registration is working as intended and the function isn't used anywhere else.

akvo/rsr/models/organisation.py
KasperBrandt: Yup, true.. Updated the functions to be similar, based on ```all_projects()```.

akvo/urls.py
KasperBrandt: The 'activate' view is directly linked to the django-registrations module, I tried to work around that since it was hard to redirect the user to a form where it can fill in a password and first and last name. That's usually already done at that point..

akvo/templates/registration/invited_user_notification_message.html
KasperBrandt: Done :+1:

akvo/templates/registration/invite_user_message.html
KasperBrandt: Done :+1:

akvo/urls.py
KasperBrandt: For now I've kept this the same (even added an employment.pk to the URL), since I'm not 100% sure the current invite process could work using the default activation link.

akvo/rest/views/user_management.py
KasperBrandt: Done :+1:. The pre- and post-save signals have been rewritten so they're more selective in sending an email. Also, approving the employment now happens after the user has filled in the form. Or, it happens immediately if the user is already approved and the user is notified that he/she has been added to an organisation (new email).

zzgvh:
Looking good! Found one small issue. When I invite someone to two organisations at the same time, it's not possible to activate both memberships. When using the first activation link, I see both organisations in My organisations, but one is Not approved, and using the second activation link does not approve that org.

PR #1962, opened 2016-01-06
[#1895] One progress bar in project editor

### LINKED ISSUE #1895: Project editor refactor and RSR / IATI split
## Test plan
As this is a large change (a complete refactor) of the project editor, it should be tested thoroughly. The following points should definitely be tested:

__Editor basics:__
- All sections should save accordingly.
- After saving, a green border is shown around the fields that have been saved. Only fields that were edited will be saved.
- If fields cannot be saved, a red border is shown around the fields.
- If the user has no permission to save the project, this should be shown in an error message _(this can be faked by editing a project and removing the user's organisation from the project)_.
- If clicked on any help icon, the help text should display.
- Fields that have a max length should show the number of characters left while the user is typing.

__Special cases:__
- Only a superuser should have the 'superuser options' on top of the screen. When these options are selected (e.g. private or impact project), the project should immediately be changed (to a private or an impact project).
- It should be possible to switch between an 'RSR related project' and a 'Related project IATI identifier' for related projects.
- It should not be possible to remove the last partner. The delete icon will then be hidden.
- It should not be possible to select multiple reporting partners.
- When the reporting partner is chosen, the 'Secondary reporter' field is shown. For all other partner roles, the 'Funding amount' field is shown. However, the 'Funding amount' field is only active when the partner is a funding partner.
- The 'Other label extra info' for budget items should only be enabled when the 'Other' budget item is selected.
- It should be possible to switch between uploading a document and specifying the URL of a document for project documents.

__Validation sets and progress bars:__
- It should be possible in the 'old' admin to set the validation sets _(see /en/admin/rsr/projecteditorvalidationset/)_. By default the RSR and IATI validation sets have been created.
- Superusers, RSR admins and organisation admins of a paying partner should be able to select or delete any validation set (called progress bar in the editor) of the project in the project editor, except for the RSR validation set.
- The RSR validation set should be selected for all projects by default and for all new projects, and it should not be possible to change any field of the RSR validation set.
- It should be possible to select a progress bar on top of the project editor and the progress bars per section should respond to the selected progress bar.
- It is possible to have 'OR'-mandatory fields, meaning that only one of multiple fields needs to be filled in for the progress bar to respond _(see start date planned and actual in the RSR validation set)_.
- Mandatory fields in a related object _(with the slightly yellow background)_ should only be taken along when one of the fields of the related object is filled.
- Related objects themselves _(e.g. 'sectors')_ can also be mandatory, indicated by the red * behind the header of the related object. This means that at least one of these related objects should be filled _(e.g. at least one sector should have a field filled in)_.
- The combination of the above rules could lead to a full progress bar when nothing is filled in, but a partly filled progress bar when the user starts to fill in a related object _(e.g. in IATI a related project is not mandatory, but when you fill in a related project, you should fill both the project and the relation fields)_.

## Issue description
For an overview of the discussion around the RSR and IATI editor split, see [this document](https://docs.google.com/document/d/1d-9ogB9WoVvHoP_sYXzLGMfByqE2xRvlBatrdn__oMU/edit).

- [x] Code refactor
- [x] Better photo and document upload (see #1760)
- [x] Improve error handling: add the fields that not saved correctly to message
- [x] Add country / latitude / longitude as mandatory fields in the 'Add new organisation' modal
- [x] No default organisation type (Plan Finland request)
- [x] Change 'Add new organisation' to 'Save' (A4A request)
- [x] Add 'Private' checkbox for superusers
- [x] One progress bar with combined validation sets
- [x] Mandatory indicators of underlying objects only show when object is filled in

KasperBrandt:
Fixes #1895

### DIFF
--- a/akvo/rest/views/benchmark.py
+++ b/akvo/rest/views/benchmark.py
@@ -17,6 +17,3 @@ class BenchmarkViewSet(PublicProjectViewSet):
     queryset = Benchmark.objects.all()
     serializer_class = BenchmarkSerializer
     filter_fields = ('project', )
-
-    def get_queryset(self, related_to='project__'):
-        return super(BenchmarkViewSet, self).get_queryset(related_to)
--- a/akvo/rest/views/benchmark_name.py
+++ b/akvo/rest/views/benchmark_name.py
@@ -16,5 +16,4 @@ class BenchmarknameViewSet(PublicProjectViewSet):
     queryset = Benchmarkname.objects.all()
     serializer_class = BenchmarknameSerializer
 
-    def get_queryset(self, related_to='benchmark__project__'):
-        return super(BenchmarknameViewSet, self).get_queryset(related_to)
+    project_relation = 'benchmark__project__'
--- a/akvo/rest/views/budget_item.py
+++ b/akvo/rest/views/budget_item.py
@@ -18,9 +18,6 @@ class BudgetItemViewSet(PublicProjectViewSet):
     serializer_class = BudgetItemSerializer
     filter_fields = ('project', 'label', 'type', )
 
-    def get_queryset(self, related_to='project__'):
-        return super(BudgetItemViewSet, self).get_queryset(related_to)
-
 
 class CountryBudgetItemViewSet(PublicProjectViewSet):
     """
@@ -28,6 +25,3 @@ class CountryBudgetItemViewSet(PublicProjectViewSet):
     queryset = CountryBudgetItem.objects.all()
     serializer_class = CountryBudgetItemSerializer
     filter_fields = ('project', 'code', )
-
-    def get_queryset(self, related_to='project__'):
-        return super(CountryBudgetItemViewSet, self).get_queryset(related_to)
--- a/akvo/rest/views/custom_field.py
+++ b/akvo/rest/views/custom_field.py
@@ -25,6 +25,3 @@ class ProjectCustomFieldViewSet(PublicProjectViewSet):
     queryset = ProjectCustomField.objects.all()
     serializer_class = ProjectCustomFieldSerializer
     filter_fields = ('project', 'name', 'max_characters', 'section', )
-
-    def get_queryset(self, related_to='project__'):
-        return super(ProjectCustomFieldViewSet, self).get_queryset(related_to)
--- a/akvo/rest/views/goal.py
+++ b/akvo/rest/views/goal.py
@@ -17,6 +17,3 @@ class GoalViewSet(PublicProjectViewSet):
     queryset = Goal.objects.all()
     serializer_class = GoalSerializer
     filter_fields = ('project', )
-
-    def get_queryset(self, related_to='project__'):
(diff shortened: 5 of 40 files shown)


zzgvh:
Progress bar looks good, however when I try to publish I get an unsaved section error:

screenshot_99

The sections in the list seem to correlate with the sections where I entered data.

zzgvh:
Some projects are seemingly ready for publishing, but can't be published. There is no indication why that is:
screenshot_100

zzgvh:
Being all Reactish stuff I think it's good if Gabe reviews. Or you give me a thorough walk-through so I understand enough to be able to make useful input regarding the client side code!

gabemart:
Code checked :+1:

KasperBrandt:
@zzgvh I can't reproduce the first comment. It is correct that this message should be displayed when there's unsaved changes, but not when you press save for each of the sections. Are you sure you've saved all sections individually?

And for the second comment, it actually looks like there's no validation set selected (the general progress bar is empty). Or at least something went wrong there, any clue how you did that?

zzgvh:
Bah, I can't reproduce any of the issues above. But I found a new one! See if you can provoke it too.

I start out with a publishable project. I open the project in a new tab, or reload the current tab. Then I immediately click Publish. The button goes "gray" and I can see the PATCH request in the terminal. But it seems the response is lacking something because the button never goes back to normal:

screenshot_103

If I now open a form tab and press Save (I haven't made any changes, I just use the Save button), the Publish button gets activated again:

screenshot_104

However the app is now in a broken state, because if I reload the page the project is actually published:

screenshot_105

The brokenness isn't critical, if I press the Publish button after having saved, nothing happens, since the project in reality already is published. But this is bound to confuse some poor user.

zzgvh:
Budgets can be negative. Should this be allowed?

zzgvh:
Publishing works perfectly now. Can't replicate the bug above.

zzgvh:
I guess you already know this, but sector percentages can't be added/edited, just viewed. Presumably they get imported via IATI documents.  However one nice-to-have would be validation of the sector codes, at least for the ones we have lists for; the three and five digit DAC lists. As it is now, you can get *Unknown code* as sector on the project page which is kinda ugly.

KasperBrandt:
- Budget items: Good point. But I think this has always been the case in RSR so far, so I think it's better to treat it in a separate item. IATI doesn't say anything about positive or negative values, although I've seen transactions that were negative.
- Sector percentages: I didn't know.. It should just be a normal field, and editable. You do have to select the IATI validation set for the field to appear though.
- Sector codes: In the current project editor, this is actually implemented (e.g. you select 3-digit vocabulary and get a list of all 3-digit codes). But since I've now made all fields generic, it was hard to fix that again (e.g. it should be possible to switch between a free text field and a dropdown).

zzgvh:
Budgets items: yep it's a separate issue. If at all. Sectors: My bad, forgot about the IATI validation.

PR #1967, opened 2016-01-07
[#1510 #1778] Refactor 'Connect to employer'

### LINKED ISSUE #1510: Cannot connect to organisations with a special character
## Test plan
1. Check if there's an organisation in the database with a special character, like "Witteveen + Bos"
2. In the MyRSR > My details section, try to connect to this organisation
3. The request should go through normally, and _organisation name (not approved)_ should appear in the organisations list

## Issue description
It's not possible to connect to organisations with a special character in the name, such as 'Witteveen + Bos'.

### LINKED ISSUE #1778: MyRSR - search short and longname
## Test plan
GIVEN the My Details page of MyRSR

WHEN trying to connect to an employer
THEN the organisation field should be red IF no organisation is not filled in
AND the organisation field should be red IF the user is already connect to the selected organisation
AND the form / button should be disabled IF the user selected a correct organisation

WHEN the request to join an employer is successful
THEN the organisation should be added to the organisations list above, with (Not approved) added

WHEN the user clicks on the red cross behind an organisation
THEN the loading icon should replace the red cross
AND the organisation should disappear IF the employment is deleted successfully

## Issue description
In MyRSR > My details > My organisations box:
When linking to employer: the search field only searches on short name. I would expect it also searches on an organisation's long name. E.g if a user wants to connect to an organisation and searches for 'Health Alive Foundation', nothing pops up. Instead I need to know upfront that I need to search for HAF, otherwise I can't find the organsation.

KasperBrandt:
- Fixes #1510 
- Fixes #1778

### DIFF
--- a/akvo/rest/views/user.py
+++ b/akvo/rest/views/user.py
@@ -101,6 +101,7 @@ def request_organisation(request, pk=None):
 
         serializer.data['country_full'] = CountrySerializer(country).data
         serializer.data['organisation_full'] = OrganisationSerializer(organisation).data
+        serializer.data['id'] = employment.pk
 
         return Response(serializer.data)
     else:
--- a/akvo/rsr/models/__init__.py
+++ b/akvo/rsr/models/__init__.py
@@ -291,7 +291,7 @@ rules.add_perm('tastypie.change_apikey', is_rsr_admin | is_org_admin | is_org_us
 
 rules.add_perm('rsr.add_employment', is_rsr_admin)
 rules.add_perm('rsr.change_employment', is_rsr_admin | is_org_admin | is_org_user_manager)
-rules.add_perm('rsr.delete_employment', is_rsr_admin | is_org_admin | is_org_user_manager)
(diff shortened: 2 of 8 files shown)


zzgvh:
Looks good from a functionality point of view. One weirdness though: when I search using the term **sudan** I get a couple of hits that don't seem relevant: *Mindanao State University (Mindanao State University - General Santos City)* and *SEK Budapest (SEK Budapest International School)*. Got an explanation for that?

Since this is an almost exclusively js/jsx PR I would recommend the code be reviewed by Gabe.

gabemart:
@zzgvh I think this is the fuzzy string matching working as intended.

`Mindanao State University (Mindanao State University - General Santos City)` is getting matched because the full string of `Short Name (Long Name)` contains the characters `s``u``d``a``n` in the correct order. 

So long as this hit is being shown below results that contain `Sudan` without characters in between (which seems to be the case in my testing` I think this behaviour is correct.

gabemart:
code checked :+1:

PR #1973, opened 2016-01-15
[#1971] Catch multiple reporting organisations scenario

### LINKED ISSUE #1971: Multiple reporting partners
## Test plan
GIVEN a project with multiple reporting partners
WHEN accessing the project page
THEN the project page should be shown instead of an error

_Note; This is hard to test, unless you have direct access to the database_

## Issue description
It shouldn't happen, but it still does: multiple reporting partners for a project. Currently RSR crashed in these cases, see http://sentry.support.akvo-ops.org/rsr/live/group/752/. We should catch this, just to be sure.

KasperBrandt:
Fixes #1971

### DIFF
--- a/akvo/rsr/models/project.py
+++ b/akvo/rsr/models/project.py
@@ -9,7 +9,7 @@ import math
 from decimal import Decimal, InvalidOperation
 
 from django.conf import settings
-from django.core.exceptions import ValidationError, ObjectDoesNotExist
+from django.core.exceptions import ValidationError, ObjectDoesNotExist, MultipleObjectsReturned
 from django.core.mail import send_mail
 from django.core.validators import MaxValueValidator, MinValueValidator
 from django.db import models
@@ -554,6 +554,11 @@ class Project(TimestampsMixin, models.Model):
                 iati_organisation_role=Partnership.IATI_REPORTING_ORGANISATION)
         except ObjectDoesNotExist:
             return None
+        except MultipleObjectsReturned:
+            # A project with multiple reporting organisations should not happen, but in practice
+            # it sometimes does unfortunately.
+            return self.partnerships.filter(
+                iati_organisation_role=Partnership.IATI_REPORTING_ORGANISATION)[0]
 
     @property
     def reporting_org(self):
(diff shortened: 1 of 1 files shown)


zzgvh:
I wonder does it make sense to check for secondary reporter status? So that if there are multiple reporting partners but only one not secondary you choose that one?

KasperBrandt:
I'll add that :+1:.. However, I think most of the time the secondary reporter field is just not filled in, so it won't make a lot of difference. 

Btw, In IATI terms, it should also not be possible to have 2 reporting orgs (e.g. 1 non-secondary and 1 secondary).

akvo/rsr/models/project.py
```diff
@@ -554,6 +554,19 @@ def reporting_partner(self):
                 iati_organisation_role=Partnership.IATI_REPORTING_ORGANISATION)
         except ObjectDoesNotExist:
             return None
+        except MultipleObjectsReturned:
+            # A project with multiple reporting organisations should not happen, but in practice
+            # it sometimes does unfortunately. In these cases we check if there's one "primary
+            # reporter" and return that. If not, we return the first reporting organisation.
+            primary_reporters = self.partnerships.filter(
+                iati_organisation_role=Partnership.IATI_REPORTING_ORGANISATION,
+                secondary_reporter='0'
```
zzgvh: Haven't tested, but I'm guessing this filter is more correct:
```python
.filter(iati_organisation_role=Partnership.IATI_REPORTING_ORGANISATION).exclude(is_secondary_report=True)
```
since it's a NullBooleanField.

zzgvh:
Yep, works. However I spotted a little niggle, not sure if it's worth going after:

screenshot_106

Here we have a project with two reporting orgs. I set the secondary status of one (which I can't set in the old admin) and can save the form.

screenshot_107

But if I edit the second reporting-org's secondary status I get complaints.

screenshot_108

And changing the first one also returns validation error. It seems like the first change to secondary reporter goes in, but after that the system won't accept any more stupid :wink:

KasperBrandt:
Yeah, there's no way around this scenario other than deleting one of the reporting organisations. Which is basically what should happen, since there's only one allowed :). But the logic in the project editor has also changed from this version (checkout the 1 progress bar PR and try again ;)).

zzgvh:
Ah, well it's such a corner case we'll leave it be.

PR #1976, opened 2016-01-15
[#1953] Always show indicator period values

### LINKED ISSUE #1953: Target and indicator values don't show on project page
## Test plan
GIVEN the display of a result on the project page

WHEN the result has no indicators
THEN only the result title should be shown

WHEN the result has a (or multiple) indicator(s)
THEN the indicators should be shown

WHEN an indicator has a (or multiple) period(s)
THEN each indicator period should appear on a new line

WHEN an indicator period is shown
THEN the dates are only shown when either the start or end date is filled in
AND the values (if present) are always shown, even though there are no dates filled in

## Issue description
If I fill in a target value of an indicator in the Project Editor section 05 Results & Indicators, the value doesn't show up on the project page. 

Screenshot target value: 2 Water facilities
screen shot 2015-12-16 at 10 55 56

Screenshot project page: Water facilities without number
screen shot 2015-12-16 at 10 56 08

KasperBrandt:
Fixes #1953

### DIFF
--- a/akvo/rsr/static/scripts-src/project-main.js
+++ b/akvo/rsr/static/scripts-src/project-main.js
@@ -13,54 +13,74 @@ var Accordion = ReactBootstrap.Accordion,
     Panel = ReactBootstrap.Panel,
     i18n;
 
-Indicator = React.createClass({displayName: 'Indicator',
-  render: function () {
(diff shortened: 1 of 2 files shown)


gabemart:
code checked :+1: looks good

PR #1978, opened 2016-01-18
[#1401] Update needs funding logic

### LINKED ISSUE #1401: Amount needs funding
## Test plan
GIVEN a project
WHEN it has less than ($/€) 1 funds needed (= total budget - total funding by funding partners)
THEN the funds needed amount is set to ($/€) 0

his is accomplished by using a management command:
```manage.py funds_needed [option] value```

GIVEN the funds_needed management command
WHEN there's no value parameter for the command to run
THEN an error is shown

GIVEN the funds_needed management command
WHEN only a value and no option is provided
THEN all projects with funds needed less than the indicated value should be printed to console

GIVEN the funds_needed management command
WHEN the ```--zero``` option is provided
THEN all projects with funds needed less than the indicated value should set their funds_needed to 0

GIVEN the funds_needed management command
WHEN the ```--complete``` option is provided
THEN all projects with funds needed less than the indicated value should set their status to 'Complete'

## Issue description
From the product meeting, it became clear that we have to look at 2 things regarding funding:

- [x] There are quite a few projects with $1 funding needed, these need to be set to completed.
- [x] How much funding is needed? This is only visible on the project funding page and on the second page of donations, but it should be apparent on the project list already, if relevant.

### PR DESCRIPTION
**Pull request by KasperBrandt**
Fixes #1401

- Added a management command to see which project have less than X funds needed, and to optionally update these values to 0 and/or set the projects to 'Complete'.
- Changed funds needed logic so that projects with funds needed less than 1 (so also negative values) will be set to 0.

### CODE REVIEW

File: akvo/rsr/management/commands/needs_funding.py [hunk 1]
```diff
@@ -0,0 +1,108 @@
...
+
+        if self.projects.count() > 0:
+            self.stdout.write('')
+            self.stdout.write('id   |funds_needed |')
+            self.stdout.write('-----|-------------|')
+
+            for p in self.projects:
+                nr_pspaces = 5 - len(str(p.pk))
+                nr_fspaces = 12 - len(str(p.funds_needed))
+                self.stdout.write(str(p.pk) + nr_pspaces * ' ' + '|' + nr_fspaces * ' ' +
```
zzgvh: Read up on string formatting (the [examples](https://docs.python.org/2/library/string.html#format-examples) say a lot) and then replace lines 52 - 55 with something approximately like:
```python
self.stdout.write(
        '{pk:{fill}{align}6}|{funds:{fill}{align}13}|'.format(pk=p.pk, funds=p.funds_needed, fill=' ', align='>'))
```

KasperBrandt: Awesome, didn't know those functions :open_mouth:

PR #2021, opened 2016-02-25
[#2013] Set Piwik URL to https by default

### LINKED ISSUE #2013: Change analytics URL to use https protocol
The current piwik analytics code use a hardcoded `http://` protocol for including the js file.

https://github.com/akvo/akvo-rsr/blob/v3.10/akvo/templates/piwik/tracking_code.html#L5

The analytics server supports https:// and we should use it to avoid an extra (301) http request.

KasperBrandt:
Fixes #2013


iperdomo:
:+1:

PR #2055, opened 2016-03-15
[#1920] IATI v2.02

### LINKED ISSUE #1920: IATI v2.02
## Test plan
GIVEN the [IATI v2.02 standard](http://iatistandard.org/202/activity-standard/iati-activities/iati-activity/)
WHEN looking at the RSR models and fields
THEN all IATI v2.02 fields should be available in RSR

GIVEN the IATI import
WHEN importing an IATI v2.02 file (e.g. [this file](https://raw.githubusercontent.com/IATI/IATI-Extra-Documentation/version-2.02/en/activity-standard/activity-standard-example-annotated.xml))
THEN all fields should be imported in RSR

GIVEN the project editor
WHEN the 'IATI' validation set has been checked
THEN all IATI v2.02 fields should be shown
AND both 'IATI' and 'IATI DGIS' validations sets should respond correctly

GIVEN the API (/v1/rest/)
WHEN looking at the models and fields
THEN all IATI v2.02 fields should be available in the API

GIVEN the IATI export
WHEN exporting a file
THEN it should be version '2.02'
AND all possible fields should be exported to IATI

GIVEN the IATI checks
WHEN performing the checks
THEN all new IATI v2.02 fields should be taken into account

## Issue description
We should upgrade to IATI v2.02 as soon as possible. The changelog for IATI v2.02 can be found here: http://iatistandard.org/201/upgrades/decimal-upgrade-to-2-02/2-02-changes/.

__Overview:__
- [x] Add new models, fields and codelists
- [x] Update the IATI import
- [x] Update the project editor
- [x] Update the API
- [x] Update the project editor validations
- [x] Update the IATI export
- [x] Update the IATI checks

__Model and field changes:__
- [x] Add humanitarian field to project and transaction
- [x] Add IATI activity ID field to partnership
- [x] Add vocabulary_uri field to recipient region, sector, policy marker, transaction sector, and transaction region
- [x] Add humanitarian scope model
- [x] Add status field to budget item
- [x] Add provider_org and receiver_org to planned disbursement
- [x] Add provider_org type and receiver_org type to transaction (= organisation's type)
- [x] Add document_date to project document
- [x] Change document category to support multiple categories (wrongfully implemented now)
- [x] Add indicator reference model
- [x] Add indicator period target and actual locations
- [x] Add indicator period target and actual dimensions
- [x] Add channel code field to CRS

KasperBrandt:
Fixes #1920


akvo/codelists/models/base_codelist.py
```diff
@@ -10,8 +10,9 @@
 
 
 class BaseCodelist(models.Model):
-    code = models.CharField(_(u'code'), max_length=100, blank=True, null=False)
-    version = models.ForeignKey('Version', verbose_name=_(u'version'), blank=False, null=False)
+    code = models.CharField(_(u'code'), max_length=100, blank=True, null=False, db_index=True)
+    version = models.ForeignKey('Version', verbose_name=_(u'version'), blank=False, null=False,
+                                db_index=True)
```
zzgvh: It looks like the **name** field could also be included in the abstract model, and possibly **description** too. Also the **__unicode__()** method might be declared here rather than in all subclasses as it looks like the implementation is the same for many if not all of them.

Maybe it even makes sense to create a few abstract base classes and use them as mixins.

akvo/iati/checks/fields/budgets.py
```diff
@@ -0,0 +1,50 @@
...
+    :param project: Project object
+    :return: All checks passed boolean, [Check results]
+    """
+    checks = []
+    all_checks_passed = True
+
+    for budget in project.budget_items.all():
+        if budget.amount is None:
+            all_checks_passed = False
+            checks.append((u'error', u'budget (id: %s) has no amount' % str(budget.pk)))
```
zzgvh: Use .format(). for string formatting. You get
  * automatic casting of non-string types
  * python3 compatibility :wink: 

(Using Pycharm's regex find/replace I think you can fix this in just a few passes, one for each type of casting)

akvo/iati/checks/fields/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    'planned_disbursements',
+    'policy_markers',
+    'related_activities',
+    'reporting_org',
+    'results',
+    'sectors',
+    'status',
+    'title',
+    'transactions',
+]
```
zzgvh: I think there's enough code duplication in the fields functions to warrant some abstraction for the field checks module. The most common check seems to be if a field is empty or not. And if it is the action is the same; set **all_checks_passed** to False and add an error string.

Sketch of a base class and an implementing class:

```python
class FieldBase(object):
    def __init__(self, project):
        self.project = project
        self.all_checks_passed = True
        self.errors = []
        self.warnings = []

    def required(self, obj, fields):
        for field in fields:
            if not getattr(obj, field, False):
                self.all_checks_passed = False
                self.errors.append(u'{} (id: {}) has no {}'.format(
                        self.__class__.__name__.lower(), obj.pk, field.replace('_', ' ')))

    def execute(self):
        for model in models:
            model.check()
            self.all_checks_passed = model.all_checks_passed or self.all_checks_passed
            self.errors += model.errors
            self.warnings += model.warnings


class Budget(FieldBase):
    def check(self):
        required_fields = ['amount', 'period_start', 'period_end', 'value_date']
        for budget in self.project.budget_items.all():
            self.required(budget, required_fields)

            if not budget.currency and not self.project.currency:
                self.all_checks_passed = False
                self.errors.append(u'budget (id: 'u'{}) has no currency and no default currency '
                                   u'specified'.format(budget.pk))
```

akvo/iati/checks/fields/crs_add.py
```diff
@@ -0,0 +1,53 @@
...
+                               str(flag.pk)))
+
+        if not crs.loan_status_year and (crs.loan_status_currency or crs.loan_status_value_date or
+                crs.interest_received is not None or crs.principal_outstanding is not None or
+                crs.principal_arrears is not None or crs.interest_arrears is not None):
+            all_checks_passed = False
+            checks.append((u'error', u'CRS (id: %s) has no loan status year specified' %
+                           str(crs.pk)))
+
+        if not (crs.loan_status_currency or project.currency) and \
```
zzgvh: Personally I loathe \\ in python code :stuck_out_tongue: It's ugly, it can usually be avoided can actually cause bugs if you accidentally add a space after the \\. [See python docs on the subject](https://docs.python.org/2/howto/doanddont.html#using-backslash-to-continue-statements) 
Here a set of () wrapping the whole if statement condition should do the job. Occurs multiple times below.

akvo/iati/checks/fields/sectors.py
```diff
@@ -0,0 +1,105 @@
...
+            for voc_key in sectors_dict.keys():
+                if len(sectors_dict[voc_key]) > 1:
+                    voc_percentage = 0
+
+                    for percentage in sectors_dict[voc_key]:
+                        try:
+                            voc_percentage += Decimal(percentage)
+                        except (ValueError, TypeError):
+                            all_checks_passed = False
+                            checks.append((u'error', u'multiple sectors with vocabulary \'%s\' '
```
zzgvh: Python has two string delimiters for the very reason that you usually don't need to escape characters by using them :stuck_out_tongue_winking_eye:

akvo/iati/exports/elements/budget.py
```diff
@@ -39,9 +47,11 @@ def budget(project):
                 value_element.attrib['currency'] = budget_item.currency
 
             if budget_item.other_extra:
-                value_element.attrib['{http://akvo.org/iati-activities}label'] = budget_item.other_extra
-            elif budget_item.label.label:
-                value_element.attrib['{http://akvo.org/iati-activities}label'] = budget_item.label.label
+                value_element.attrib['{http://akvo.org/iati-activities}label'] = budget_item.\
```
zzgvh: To avoid the horrid \\:
```python
    akvo_ns_label = '{http://akvo.org/iati-activities}label'
    ...
    value_element.attrib[akvo_ns_label] = budget_item.other_extra
```

akvo/iati/exports/elements/capital_spend.py
```diff
@@ -14,7 +14,7 @@ def capital_spend(project):
     :param project: Project object
     :return: A list of Etree elements
     """
-    if project.capital_spend_percentage:
+    if project.capital_spend_percentage == 0 or project.capital_spend_percentage:
```
zzgvh: I think a better expression here is ```if project.capital_spend_percentage is not None:``` Found multiple times below.

akvo/iati/exports/elements/conditions.py
```diff
@@ -22,9 +22,10 @@ def conditions(project):
         conditions_element.attrib['attached'] = '0'
 
     for condition in project.conditions.all():
-        if condition.type:
-            condition_element = etree.SubElement(conditions_element, "condition")
-            condition_element.attrib['type'] = condition.type
+        if condition.type or condition.text:
```
zzgvh: This leads to a possible error on line 34 if condition.type is False, since condition_element then won't be assigned to.

akvo/iati/exports/elements/country_budget_items.py
```diff
@@ -16,22 +16,24 @@ def country_budget_items(project):
...
+                if budget_item.code:
+                    element = etree.SubElement(country_budget_items_element, "budget-item")
+                    element.attrib['code'] = budget_item.code
 
-    for budget_item in project.country_budget_items.all():
-        if budget_item.code:
-            element = etree.SubElement(country_budget_items_element, "budget-item")
-            element.attrib['code'] = budget_item.code
+                if budget_item.description:
+                    description_element = etree.SubElement(element, "description")
```
zzgvh: element not guaranteed to have been assigned here.

akvo/iati/exports/elements/country_budget_items.py
```diff
@@ -16,22 +16,24 @@ def country_budget_items(project):
...
+                    narrative_element = etree.SubElement(description_element, "narrative")
+                    narrative_element.text = budget_item.description
 
-            if budget_item.description:
-                description_element = etree.SubElement(element, "description")
-                narrative_element = etree.SubElement(description_element, "narrative")
-                narrative_element.text = budget_item.description
+                country_budget_items_element.append(element)
```
zzgvh: element not guaranteed to have been assigned here.

akvo/iati/exports/elements/result.py
```diff
@@ -17,86 +17,156 @@ def result(project):
...
-                        narrative_element = etree.SubElement(comment_element, "narrative")
-                        narrative_element.text = period.actual_comment
-
-                indicator_element.append(period_element)
-
-            element.append(indicator_element)
-
-        result_elements.append(element)
+            for indicator in res.indicators.all():
+                if indicator.measure or indicator.ascending is not None or indicator.title or \
```
zzgvh: Now I've seen this kind of really long ifs a number of times and they're hard to read. A helper function along these lines maybe:

```python
def at_least_one_value_field(obj, fields):
    "Returns True if at least one field on obj has a truthy value, otherwise returns False"
    for attr in attrs:
        if getattr(obj, attr, None):
            return True
    return False

fields = ['measure', 'title', 'description', 'baseline_year', 'baseline_value', 'baseline_comment']
if (at_least_one_value_field(indicator, fields) or indicator.ascending is not None
           or indicator.references.all() or indicator.periods.all()):
...
```

Also note that by having the ORM queries last, there's a good chance they never have to be checked and they're the most expensive by far. (Maybe this is too much nit-picking? I can't decide, but those 4-line if statements are ugly!)

akvo/iati/exports/elements/transaction.py
```diff
@@ -88,74 +93,94 @@ def transaction(project):
     transaction_elements = []
 
     for trans in project.transactions.all():
-        element = etree.Element("transaction")
+        if trans.reference or trans.humanitarian is not None or trans.transaction_type or \
```
zzgvh: No...this has to go!

akvo/iati/imports/__init__.py
```diff
@@ -291,7 +293,7 @@ def get_attrib(self, element, attr, field, default=''):
 
     def get_attrib_as_int(self, element, attr, field, default=None):
         value = self.get_attrib(element, attr, field, default)
-        if value:
+        if value == '0' or value:
```
zzgvh: Not needed. bool('0') == True.

Issue #2130, opened 2016-04-29
Bring back general 'Add update' button


KasperBrandt:
After various mentions that there is no more 'Add update' button on top, we should actually bring it back. Also because the big restructuring of RSR is not planned anymore.

KasperBrandt:
Fixed in #2131

PR #2131, opened 2016-05-02
[#2130] Fixed

loicsans:


KasperBrandt:
It took me quite a while to grasp this, but code seems ok 👍

PR #2202, opened 2016-05-23
[#2196] Invite content owned organisations

### LINKED ISSUE #2196: Inviting and accepting users of content owned organisations
The ```organisation.content_owner``` field is used in some parts of RSR, but not fully yet. This is the most obvious in the "invite user" functionality (User management in MyRSR) and when approving a user of a content owned organisation.

- [x] In the 'invite users' dropdown, the organisations of which the user's organisation is content owner should also be populated in the list.
- [x] Once a user requests to join an organisation, the users of the content owner (and their content owner, etc) should also receive an email. But only when they're Admin or User manager and when they're user account has 'is_support' set to True.

damienallen:
Fixes #2196

### DIFF
--- a/akvo/rsr/signals.py
+++ b/akvo/rsr/signals.py
@@ -272,6 +272,11 @@ def employment_post_save(sender, **kwargs):
                     employers__organisation=organisation,
                     employers__group__in=[user_managers_group, admins_group],
                     is_support=True
+                ) |
+                active_users.filter(
+                    employers__organisation=organisation.content_owner,
+                    employers__group__in=[user_managers_group, admins_group],
+                    is_support=True
                 )
             ).distinct()
             rsr_send_mail_to_users(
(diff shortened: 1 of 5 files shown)


akvo/rsr/signals.py
```diff
@@ -272,6 +272,11 @@ def employment_post_save(sender, **kwargs):
                     employers__organisation=organisation,
```
KasperBrandt: You could also rewrite this as `employers__organisation__in=[organisation, organisation.content_owner]`.

KasperBrandt:
I've changed it around a little, including a 'new' definition of content owned organisations which was already partly implemented:

An organisation can be content owned by setting the 'content_owner' organisation in the organisation admin. In addition, if the organisation is a paying partner, then it content owns all implementing partners of all projects they participate in. Except when the implementing partner is a paying organisation itself.

With this in mind, could you check the second test plan @damienallen? See https://github.com/akvo/akvo-rsr/issues/2196#issuecomment-221491491.

PR #2206, opened 2016-05-26
[#1804] Switch to IATI countries

KasperBrandt:
Fixes #1804

Issue #1804, opened 2015-09-16
Move to IATI country list by default


lauraroverts:
We now have our own country list in RSR, and IATI also uses a country list (which we also use for some fields, such as recipient country). I think we should completely switch to the IATI country list, but that also involves some other changes (filters on RSR pages, etc).

- [x] Projects
  - [x] Remove the rsr_projectlocation.country field from the project editor
  - [x] Remove all mentions of rsr_projectlocation.country in the validation sets
  - [x] Do not hide recipient countries and regions in the RSR validation set
  - [x] Convert any country of a project location to a recipient country
  - [x] Remove the country field as a mandatory field for publishing in RSR
  - [x] Update project page
    - [x] Project header
    - [x] Full report tab (remove country field from location)
    - [x] Donation thanks page
  - [x] Update projects list
    - [x] Update mention of country per project
    - [x] Update location filter
  - [x] Update IATI import
  - [x] Add comment to state that location.country is legacy
- [x] Project updates
  - [x] Remove project update location countries (not used)
  - [x] Remove location filter on updates list
- [x] Organisations
  - [x] Add new iati_country field
  - [x] Convert existing countries to IATI countries
  - [x] Replace country field in 'old' organisation admin
  - [x] Add comment to state that location.country is legacy
  - [x] Update location filter on organisations list
  - [x] Update organisations list
- [x] Employments
  - [x] Convert the employment countries to IATI countries
  - [x] Update 'Request to join an organisation' module
- [x] Make all migrations reversable

stellanl:
Don't forget the API.
Bumping the modification date for projects and updates will make them reload, but Up currently assumes the country list will almost never change. If all the countries suddenly have different IDs it will be very confusing. Maybe we need a mechanism in the API for indicating an invalidation of cached data.

KasperBrandt:
@stellanl I'm planning the following changes, let me know if you are ok with that:

- The _country_ field for a project location stays in the model and API as it is right now. Basically, as a legacy field.
- The location field gets removed from the project editor and all countries are mapped to a _recipient country_.
- All filters etc are based on the _recipient country_ and _recipient region_ fields.

stellanl:
Should be ok, I only display the country, so the old Up apps should be ok as long as the field stays around. It is maybe an hour's work to switch to a new country field for some an Up release

stellanl:
There is also a country id field in the employment application API.

KasperBrandt:
## Test plan
To do.

[...]

PR #2207, opened 2016-05-26
[#2189] Move to IATI statuses

### LINKED ISSUE #2189: Move to IATI statuses
This is probably a bigger issue than we expect, but we should replace the current project.status with the IATI entries. A mapping will be done like this:

```
STATUS_TO_CODE = {
    'N': '6',
    'H': '1',
    'A': '2',
    'C': '3',
    'L': '5',
    'R': '6',
}
```

- [x] Convert status codes
- [x] Update IATI import / export
- [x] Update project editor
- [x] Update project page and project entry in projects list
- [x] Cancelled and completed projects can't be updated (see https://github.com/akvo/akvo-rsr/issues/2006#issuecomment-218499657)
- [x] Check where else in RSR the project's status is used

damienallen:
Fixes #2189


damienallen:
Watch for conflicting migration ids with #1804 before merging.

akvo/iati/imports/mappers/defaults.py
```diff
@@ -181,11 +181,9 @@ def do_import(self):
         status = self.get_child_elem_attrib(
                 self.parent_elem, 'activity-status', 'code', 'status')
 
-        if status in CODE_TO_STATUS.keys():
-            status = CODE_TO_STATUS[status]
-        else:
+        if not status:
             self.add_log('activity-status@code', 'status', 'invalid status code')
-            status = Project.STATUS_NONE
+            status = 0
```
KasperBrandt: This should probably be a string instead of integer, right?

akvo/iati/imports/mappers/defaults.py
```diff
@@ -181,11 +181,9 @@ def do_import(self):
...
 
         return self.update_project_field('status', status)
```
KasperBrandt: Update 'status' (this refers to the ```project.status``` field) to 'iati_status' here.

akvo/rsr/migrations/0072_auto_20160525_1008.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('rsr', '0071_auto_20160525_1006'),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_iati_status),
+        migrations.RunSQL(
```
KasperBrandt: This gives me an error. Might be easier to just use a Python script (or see the Django 1.9 docs for examples of RunSQL).

akvo/templates/organisation_main.html
```diff
@@ -123,47 +123,67 @@ <h4 class="topMargin">{% trans 'IATI file' %}</h4>
             </a>
           </h2>
           <dl class="dl-horizontal">
-            <dt>{% trans 'Active' %}</dt>
+            <dt class="iatiStatuses">{% trans 'Implementation' %}</dt>
             <dd>
               {% if organisation.published_projects.status_active.count %}
-                <a href="{% url 'project-directory' %}?organisation={{ organisation.pk }}&amp;status=A">
+                <a href="{% url 'project-directory' %}?organisation={{ organisation.pk }}&amp;iati_status=2">
```
KasperBrandt: In this list I would put status 1 (Pipeline/identification) first, then status 2 (Implementation).

akvo/rsr/models/project.py
```diff
@@ -92,27 +92,54 @@ class Project(TimestampsMixin, models.Model):
...
-                    u'not yet started.<br/>'
-                    u'2) Active: the implementation phase has begun.<br/>'
-                    u'3) Completed: the project has been completed.<br/>'
-                    u'4) Cancelled: the project never took place or work stopped before it was '
-                    u'fully implemented.<br/>'
-                    u'5) Archived: projects are archived when the reporting partner no longer uses '
-                    u'RSR.')
+    status = ValidXMLCharField(_(u'status'), max_length=1, choices=STATUSES, db_index=True, default=STATUS_NONE)
+    iati_status = ValidXMLCharField(
+        _(u'iati status'), max_length=1, choices=([('0', '')] + codelist_choices(ACTIVITY_STATUS)),
```
KasperBrandt: Might be less confusing for users to name this field the same (u'status' instead of u'iati status'). In the project editor they'll otherwise see an IATI reference whilst that is not really necessary. Also update this in the migration btw.

akvo/rsr/models/project.py
```diff
@@ -92,27 +92,54 @@ class Project(TimestampsMixin, models.Model):
...
+        db_index=True, default='0',
+        help_text=_(u'There are six different project statuses:<br/>'
+                    u'1) Pipeline/identification: the project is being scoped or planned<br/>'
+                    u'2) Implementation: the project is currently being implemented<br/>'
+                    u'3) Completion: the project is complete or the final disbursement has been made<br/>'
+                    u'4) Post-completion: the project is complete or the final disbursement has been made, '
+                    u'but the activity remains open pending financial sign off or M&E<br/>'
```
KasperBrandt: ".. but the activity .." --> ".. but the project ..". Also change this in the migration.

KasperBrandt:
Also change the field in the 'old' project admin to use ```iati_status``` instead of ```status```.

damienallen:
**Only merge after #2206 has been merged**

PR #2209, opened 2016-05-27
[#2003] Management command to clean up empty projects

### LINKED ISSUE #2003: Empty projects
Similar to #1753, we should add a management command for deleting empty and stale projects. 

We should remove all non-published projects without any data (except a partnership, which will usually be present) older than a week.

damienallen:
Fixes #2003


KasperBrandt:
@damienallen I did the management command a bit differently. Could you check out the code and see if it still works :)?

PR #2212, opened 2016-05-31
#1801 Move to IATI currencies

### LINKED ISSUE #1801: Move to IATI currencies
Currently, there's only 2 options for a currency in RSR: Euros and US Dollars. However, with the introduction of IATI fields we have also incorporated the IATI codelists for some fields that allow the list of IATI currencies including all possible currencies. We should extend the currencies of RSR, keeping in mind:

- [x] Project editor should allow to select all possible currencies and display it for any 'currency' field.
- [x] Show the currency code (e.g. EUR, USD, YEN) everywhere:
  - [x] Projects list
  - [x] Project page (summary, full report and finance tabs)
  - [x] Organisation page
- [x] All fields in the project editor should respond to the currency selection. When you switch to a different currency, this should happen for all fields that use a currency in the editor.

damienallen:
Fixes #1801


akvo/rsr/models/budget_item.py
```diff
@@ -119,11 +122,20 @@ def get_label(self):
         else:
             return self.__unicode__()
 
+    def get_currency(self):
```
KasperBrandt: Could be rewritten to:
`return self.currency if self.currency else self.project.currency`

akvo/rsr/models/organisation.py
```diff
@@ -456,6 +457,25 @@ def iati_file(self):
 
     # New API
 
+    def has_multiple_project_currencies(self):
```
KasperBrandt: Could be rewritten to:
`return self.published_projects().distinct().count() != self.org_currency_projects_count()`

akvo/rsr/models/project.py
```diff
@@ -980,6 +990,41 @@ def external_links(self):
     def budget_total(self):
         return Project.objects.budget_total().get(pk=self.pk).budget_total
 
+    def has_multiple_budget_currencies(self):
+        budget_items = BudgetItem.objects.filter(project__id=self.pk)
+        num_currencies = len(set([c.currency if c.currency else self.currency for c in budget_items]))
+
+        if num_currencies > 1:
```
KasperBrandt: Could be rewritten to:
`return num_currencies > 1`

akvo/rsr/models/transaction.py
```diff
@@ -162,8 +162,17 @@ def receiver_organisation_show_link(self):
                                                    self.receiver_organisation.name)
         return ''
 
+    def get_currency(self):
```
KasperBrandt: Could be rewritten to:
`return self.currency if self.currency else self.project.currency`

Issue #1989, opened 2016-02-02
Add explanation user roles in User Management


lauraroverts:
User roles and permissions are not clear for all RSR users. We need to add an explanatory text to indicate the different user roles and permissions. Or link to http://rsrsupport.akvo.org/article/show/user-roles-and-permissions 

![screen shot 2016-03-01 at 09 44 58](https://cloud.githubusercontent.com/assets/4179989/13422011/4987c844-df92-11e5-83d8-ff637f89fa9d.png)

KasperBrandt:
We'll have to look into this to see if it is still necessary after the updated user roles and permissions in #2115.

AkvoAnt:
A link to the User Management support page is now in place.

PR #2236, opened 2016-06-07
[#2196] Updated permissions on user management page

### LINKED ISSUE #2196: Inviting and accepting users of content owned organisations

KasperBrandt:
Fixes #2196

### DIFF
--- a/akvo/rsr/static/scripts-src/my-user-management.js
+++ b/akvo/rsr/static/scripts-src/my-user-management.js
@@ -10,7 +10,10 @@ var Button,
     Modal,
     SplitButton,
     Table,
-    initial_data,
+    initial_employment_data,
+    orgAdmin,
+    organisation_data,
+    role_data,
     i18n;
 
 // CSRF TOKEN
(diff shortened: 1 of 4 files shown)


damienallen:
### Issues

- [ ] M&E managers lack access to User Management
- [ ] User managers (and probably M&E managers) cannot invite/change role to 'Admins' as intended _but_ can demote Admins to other user groups

KasperBrandt:
Just checked the overview in https://github.com/akvo/akvo-rsr/issues/1989, but it's actually correct that M&E managers cannot access user management, my bad 😄 .

PR #2247, opened 2016-06-15
[#1956] Result and indicator ordering

### LINKED ISSUE #1956: Random order indicators
- [x] Add an `order` field for results and for indicators.
- [x] It should be possible to set the order of results and indicators in the project editor. Preferable by drag and drop, otherwise by 'Move up' and 'Move down'.
- [x] The results framework (both My results and on project page, but that's the same code) should adhere to this ordering.
- [x] Preserve ordering for imported child results and indicators
- [x] Initial ordering should match order of entry

I created two test projects: 
- Parent http://rsr.test.akvo.org/en/project/4239/
- Child http://rsr.test.akvo.org/en/myrsr/project_editor/4240/ 

In 05 Results and Indicators I started to create a results framework with one result and multiple indicators. I named the indicators Indicator 1, 2 and 3. After I save it and go back to that section, the indicators appear in a random order (2-3-1) in the backend and frontend: 

screen shot 2015-12-22 at 09 48 17

screen shot 2015-12-22 at 09 45 58

If I now import my results framework at my child project, the imported result and indicators do have a different order again (1-3-2): 
screen shot 2015-12-22 at 09 42 47

This is going to be complicated and confusing when I add a results framework with multiple results and indicators.

I would expect the same order/structure of how I enter my framework and after importing or viewing in the frontend. I also want to be able to change the order with the ability to move results and indicators up and down.

damienallen:
Fixes #1956


akvo/rest/views/project_editor.py
```diff
@@ -572,6 +573,67 @@ def project_editor(request, pk=None):
...
+    else:
+        errors += ['Invalid item type']
+
+    # assign order if it doesn't already exist
+    if not item_list[0].order:
+        for i, item in enumerate(item_list):
+            item.order = i
+            item.save()
+
+    item_original_order = Indicator.objects.get(id=item_id).order
```
KasperBrandt: A check for ```item_type``` is missing here. For results, the ordering now does not work.

akvo/rest/views/project_editor.py
```diff
@@ -572,6 +573,67 @@ def project_editor(request, pk=None):
...
+        item_selected = Result.objects.get(id=item_id)
+        item_list = Result.objects.filter(project_id=project_pk)
+    elif item_type == 'indicator':
+        item_selected = Indicator.objects.get(id=item_id)
+        item_list = Indicator.objects.filter(result_id=item_selected.result_id)
+    else:
+        errors += ['Invalid item type']
+
+    # assign order if it doesn't already exist
+    if not item_list[0].order:
```
KasperBrandt: If ```item_type``` is not 'result' or 'indicator', then ```item_list``` is not instantiated here, which will give a ```NameError```.

akvo/rest/views/project_editor.py
```diff
@@ -572,6 +573,67 @@ def project_editor(request, pk=None):
...
+    item_original_order = Indicator.objects.get(id=item_id).order
+
+    if item_direction == 'up' and not item_original_order < 1:
+        item_swap = item_list.get(order=item_original_order - 1)
+        item_swap.order = item_original_order
+        item_swap.save()
+
+        swap_id = item_swap.id
+
+        item_selected.order = item_original_order-1
```
KasperBrandt: Same here, if ```item_type``` is not 'result' or 'indicator', then ```item_selected``` is not instantiated.

KasperBrandt:
Gives an error:
1. Create a new project
2. Add 3 new results and save
3. Try to move the first item down

It then sends this to the server:
```
item_type:result
item_id:1576_new-0
item_direction:down
```

The item_id is incorrect, and the result node has ```id="result.516"```, so I'm not sure where it gets the id from.

PR #2282, opened 2016-06-30
[#2270] Remove unpublished and private projects from hierarchy

### LINKED ISSUE #2270: Remove unpublished projects from the hierarchy
On the hierarchy page in RSR, we now show any project and thus also unpublished projects. Only published and public (non-private) projects should be shown.

damienallen:
Fixes #2270


KasperBrandt:
We should also update the text on the project page (summary tab), which now still also displays the numbers including unpublished/private projects:

screen shot 2016-06-30 at 16 33 54

KasperBrandt:
@damienallen Changed the PR a little, updated now.

PR #2284, opened 2016-06-30
[#2262] Remove Tastypie and convert old links to DRF

### LINKED ISSUE #2262: Look into performance of Tastypie
We should look into the Tastypie API and see whether it causes some of RSR's performance issues.

damienallen:
Fixes #2262 

To enable Non-HTML Django Debug Toolbar, add function to MIDDLEWARE_CLASSES in 66_local.conf:

```
MIDDLEWARE_CLASSES = (
    ('debug_toolbar.middleware.DebugToolbarMiddleware',
    'akvo.rsr.middleware.NonHtmlDebugToolbarMiddleware',)
    + MIDDLEWARE_CLASSES)
```


damienallen:
**INCOMPLETE DO NOT MERGE**

KasperBrandt:
Updated this PR, could you check @damienallen?

damienallen:
Project and user fields used to include URI in old API, but now only return ID, ex:

Old API: `/api/v1/project/5/`
New API: `5`

For legacy purposes, the URI should be returned.

damienallen:
Redirect `depth=1` to `/project_extra/` and create a custom endpoint for `depth=2`.

damienallen:
`?format=html` doesn't work for `/api/v1/right_now_in_akvo/` or `/api/v1/project_update/` returning `"detail": "not found"` but HTML support isn't needed anyways.

damienallen:
The redirects for `depth=1` and `depth=2` don't take `project=` into account. Thus multiple objects are returned even if the original query specified a project.

KasperBrandt:
About https://github.com/akvo/akvo-rsr/pull/2284#issuecomment-232296968: the ```format=html``` doesn't work anywhere 😄 . It's only when the format parameter is left out that it shows the HTML template.

PR #2286, opened 2016-06-30
[#2226] Hide delete button for partnerships if user would have no edit rights afterwards

### LINKED ISSUE #2226: Warn the user when they lose editing rights of project
When removing an organisation from a project, it is possible that the user loses the right to edit the project altogether. In this case, we should warn the user that they will lose the editing rights.

KasperBrandt:
Fixes #2226


damienallen:
Only hidden for admins and M&E managers, not project admins.

PR #2289, opened 2016-07-04
[#2273] Handle corrupt images in API calls

### LINKED ISSUE #2273: Internal server error for API call
This URL currently gives an internal server error: http://rsr.akvo.org/rest/v1/project_extra/787/.json

Part of the stacktrace:
```
IOError Root Cause
image file is truncated (48 bytes not processed)

...
akvo/rest/fields.py ? in to_native
    default_thumb = get_thumbnail(value, default_width, quality=99)
sorl/thumbnail/shortcuts.py ? in get_thumbnail
    return default.backend.get_thumbnail(file_, geometry_string, **options)
sorl/thumbnail/base.py ? in get_thumbnail
    thumbnail)
sorl/thumbnail/base.py ? in _create_thumbnail
    image = default.engine.create(source_image, geometry, options)
sorl/thumbnail/engines/base.py ? in create
    image = self.colorspace(image, geometry, options)
sorl/thumbnail/engines/base.py ? in colorspace
    return self._colorspace(image, colorspace)
sorl/thumbnail/engines/pil_engine.py ? in _colorspace
    return image.convert('RGB')
PIL/Image.py ? in convert
    self.load()
PIL/ImageFile.py ? in load
    "(%d bytes not processed)" % len(b))
```

damienallen:
Fixes #2273

### DIFF
--- a/akvo/rest/fields.py
+++ b/akvo/rest/fields.py
@@ -126,12 +126,18 @@ class Base64ImageField(ImageField):
                         return get_thumbnail(value, '{}'.format(max_size), quality=99)
                 except ThumbnailParseError:
                     return None
+                except IOError:
+                    return None
             # no size specification matching the name found; give up
             return None
 
         if value:
             default_width = '191' # width of update images on akvo.org/seeithappen
-            default_thumb = get_thumbnail(value, default_width, quality=99)
+            try:
+                default_thumb = get_thumbnail(value, default_width, quality=99)
+            except IOError:
+                return None
+
             try:
                 request = self.context['request']
             except KeyError:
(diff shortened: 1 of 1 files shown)


akvo/rest/fields.py
```diff
@@ -126,14 +126,18 @@ def get_thumb(request, name):
                         return get_thumbnail(value, '{}'.format(max_size), quality=99)
                 except ThumbnailParseError:
                     return None
+                except IOError:
```
KasperBrandt: These (and below as well) can be changed to:

```
except (ThumbnailParseError, IOError):
    ...
```

PR #2304, opened 2016-07-17
[#2181] Average aggregation of percentages

### LINKED ISSUE #2181: Average of indicator percentages in parent project

KasperBrandt:
Fixes #2181

### DIFF
--- a/akvo/rsr/models/indicator.py
+++ b/akvo/rsr/models/indicator.py
@@ -803,7 +803,7 @@ class IndicatorPeriodData(TimestampsMixin, models.Model):
             raise ValidationError(validation_errors)
 
         # Don't allow a data update to an aggregated parent period with 'percentage' as measurement
-        if self.period.indicator.children_aggregate_percentage():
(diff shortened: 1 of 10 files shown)


akvo/rsr/models/indicator.py
```diff
@@ -778,6 +802,13 @@ def clean(self):
                                                     u'data to it'))
             raise ValidationError(validation_errors)
 
+        # Don't allow a data update to an aggregated parent period with 'percentage' as measurement
+        if self.period.indicator.children_aggregate_percentage():
```
damienallen: Raises a TypeError, I think it should be `self.period.indicator.children_aggregate_percentage` since the `@property` decorator is used on the method.

damienallen:
Percentage values seem to be aggregated regardless of aggregation toggles (parent and child level).

damienallen:
Aggregated percentage doesn't seem to update when indicator period updates are edited at child level.

Issue #2347, opened 2016-08-26
URL for my results is incorrect


nadiagorchakova:
The issue was found in UAT RSR environment only.
Steps to reproduce:

- Go to My RSR and pick a project

- Go to My results tab and select one of the indicators on the right

- Go back two steps to project search

The search does not return any results anymore and the url looks like that: http://rsr.uat.akvo.org/en/myrsr/results/?4001/4001/4001/

zzgvh:
@gabemart , I'm assigning this to you. I've done some research and to me it looks like the bug is so far down the React rabbithole that I'd never get out if I jumped in. If you like I can show you my findings and maybe you can pick up the thread if you think it's solvable.

gabemart:
With help of @zzgvh I diagnosed the problem as the <form> element the typeahead was wrapped in. Hitting "Enter" on the typeahead input was triggering a form submit, which reloaded the page with a `?` parameter, breaking further navigation.

Simple replacing the <form> with a <div> fixes the issue.

zzgvh:
Fixed in https://github.com/akvo/akvo-rsr/pull/2352

PR #2352, opened 2016-09-05
[#2347] Change <form> to <div> to prevent useless submit

gabemart:
#2347


zzgvh:
Yep, looks good, I can't provoke the posting of the page without a valid search term.

Issue #2351, opened 2016-09-01
Mandatory fields in Project Focus (My RSR Project Editor)


nadiagorchakova:
I encountered the following issue in RSR Test env:

1. Fields 'Policy Marker' and 'Policy Marker Significance' fields are not marked as mandatory until you fill them in.
2. Even after you fill in all mandatory fields, the progress bar for the section stays yellow.

This is true when IATI or DGIS standard was checked in the Project settings.

![screen shot 2016-09-01 at 12 26 55](https://cloud.githubusercontent.com/assets/21127166/18164497/70001a6e-7040-11e6-998f-5240143a477c.png)

Please review. Thanks!

nadiagorchakova:
Same thing in RSR UAT env.
You can have a look at these (unpublished) projects:
RSR Test: http://rsr.test.akvo.org/en/myrsr/project_editor/4521/
RSR UAT: http://rsr.uat.akvo.org/en/myrsr/project_editor/4005/

nadiagorchakova:
Annabelle found same issue in RSR Live (Watershed)
http://rsr.akvo.org/en/myrsr/project_editor/4367/

zzgvh:
Looking both [here](http://iatistandard.org/202/activity-standard/overview/classifications/) and [here](http://iatistandard.org/202/rulesets/standard-ruleset/) I cannot find that Policy marker should be mandatory. Sector seems to be the only classification that's required. But I've only done a quick read. Or do we have special rule sets that go above the IATI requirements?

nadiagorchakova:
Gabriel, you're right! Geert has just changed the settings in http://rsr.test.akvo.org/en/admin/, so now only sector is mandatory. The completion bar is yellow, not matter how many fields you enter though.

gabemart:
After having a look, I agree with theory of @zzgvh that this behaviour is caused by `rsr_policymarker.policy` and `rsr_policymarker.significance` being set to "read only" in the admin interface

![image](https://cloud.githubusercontent.com/assets/2589977/18251091/ff0c6e24-737e-11e6-8af5-f450fbf411d9.png)

This status is not implemented in the project editor, or in the `project_editor.py` [template tag](https://github.com/akvo/akvo-rsr/blob/develop/akvo/rsr/templatetags/project_editor.py#L213). `mandatory_or_hidden` only checks for `validation_action` values of "1" or "2", but "readonly" sets a validation action of "3". 

[...]

PR #2357, opened 2016-09-08
#2351 mandatory fields accounting bug

zzgvh:
# Test plan

Check that a project in the project editor can be published when all mandatory fields are filled in. Especially check the behaviour in section 8, since that's where the bug occurred. However the fix might have messed up other sections 😛 Check with the RSR validation only and both the RSR and IATI validation sets.

For everything to work the validation set for IATI also needs to be fixed: A **rsr_policymarker** validation needs to be added, set to Mandatory, and the validations for **rsr_policymarker.policy_marker** and **rsr_policymarker.significance** should also be set to Mandatory if they're not: 

screenshot_124


akvo/rsr/models/project_editor_validation.py
```diff
@@ -49,14 +49,14 @@ class ProjectEditorValidation(Model):
     """
     MANDATORY_ACTION = 1
     HIDDEN_ACTION = 2
-    READ_ONLY_ACTION = 3
+    # READ_ONLY_ACTION = 3
```
punchagan: @zzgvh Is there a reason you commented this out, instead of removing the code?

punchagan:
The discussed problem is fixed and I didn't notice any new issues. 
LGTM.

akvo/rsr/models/project_editor_validation.py
zzgvh: I think we want to implement this option, so it's there as a reminder.

punchagan: I'd vote for an issue or a roadmap card with a link to this commit as a better reminder.  Sorry for the nit-pick!

akvo/rsr/templatetags/project_editor.py
```diff
@@ -214,7 +216,7 @@ def mandatory_or_hidden(validations, field):
     """
     Retrieves the mandatory and hidden fields for project editor validations.
 
-    :returns "mandatory-1 mandatory-or-2-subtitle hidden-3"
+    :returns A string on the form mandatory-{validation ID} and/or hidden-{validation_ID}
```
punchagan: Minor typo - s/string on/string of/

akvo/rsr/models/project_editor_validation.py
zzgvh: Hehe, no worries, you're right!

punchagan: Thanks!

zzgvh:
A note for the deployment of this fix. Check that there are no validations set to Read-only. If there are any they should be removed before deploying, as they will break the admin page for project editor validation sets.

punchagan:
#2372 has the missing migration file for this fix.

PR #2360, opened 2016-09-13
Fix problems with using react-datepicker for date fields

### LINKED ISSUE #2359: Date pickers don't work correctly for mandatory fields
Found this issue with http://rsr.test.akvo.org/en/myrsr/project_editor/4521/

- Date fields were the last mandatory fields that were required to be filled.  
- Even after dates were selected, the progress bar for the section did not turn green
mon sep 12 14 19 44 ist 2016

- To make it green, another field had to be filled in. (in this case currency was selected)
mon sep 12 14 20 18 ist 2016

- But, note that the date fields which were previously mandatory, are **no longer mandatory**.

punchagan:
Fixes #2359.

# Test Plan

1. Check that filling in an empty, mandatory date field in the Project editor, updates the corresponding section completion progress bar. 

2. Make sure that the mandatory * doesn't go away after the field has been filled in. 

3. Check that mandatory OR datefields work correctly.  For instance 01. General Information > Start date (actual)/(planned) with IATI validation.

### DIFF
--- a/akvo/rsr/static/scripts-src/project-editor.js
+++ b/akvo/rsr/static/scripts-src/project-editor.js
@@ -3238,7 +3238,7 @@ function setDatepickers() {
                             selected: this.state.initialDate,
                             onChange: this.handleDateChange,
                             todayButton: defaultValues.today,
-                            className: 'form-control'
+                            className: this.props.classNames
                         })
                     );
                 } else {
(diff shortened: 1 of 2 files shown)


punchagan:
@nadiagorchakova I used this bug as a way to get familiar with React and the Project Editor code, even though this is a fairly newly found bug and hasn't yet been scheduled.

PR #2372, opened 2016-09-20
[Fixes #2371] Sort mappers_prefix choices

### LINKED ISSUE #2371: Unnecessary migrations are created for IatiImport.mapper_prefix
The choices for the `mapper_prefix` field are populated using a function that lists the sub-packages of `akvo.iati.imports.mappers`.  But, since the function does a directory listing, the order of packages can change and cause a new migration to be required. 

The function should return the sub-packages sorted, to fix this.

punchagan:
The choices are obtained by listing subpackages of
`akvo.iati.imports.mappers`, and not sorting them can create unnecessary
migrations.

Also, add a couple of missing migration files from previous changes.

### DIFF
--- a/akvo/codelists/migrations/0004_auto_20160920_1045.py
+++ b/akvo/codelists/migrations/0004_auto_20160920_1045.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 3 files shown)


punchagan:
## Test Plan

Running `makemigrations` should not produce any more new migrations or utmost one, with the choices for `IatiImport.mapper_prefix` sorted.

zzgvh:
Interesting. Did you spot this when you created the missing migrations?

punchagan:
Yes, there was one extra migration showing up, but I was finding it difficult to track it back to the PR in which this went missing.  Eventually, found the bug.

PR #2376, opened 2016-09-21
Remove donations feature from RSR

### LINKED ISSUE #1885: Remove donations (estimate: 4)
The idea is to remove the donation functionality from RSR. However, this will have some impact on partners that use this feature a lot. Therefore, we need to look at what we will replace this functionality with. For example, there might be 3rd party providers that provide an easy payment / crowdfunding plugin that could be used. This needs to be investigated.

punchagan:
Fixes #1885 

# Test Plan

@nadiagorchakova gave a good test plan in [this comment](https://github.com/akvo/akvo-rsr/issues/1885#issuecomment-247945994).  Copying it over for convenience. 

- [ ] I believe the donation button and arrangement with Akvo bank account was partially deactivated (in Live) already. Needs to be checked again and fully deactivated.

- [ ] In the Project Editor please rename field 'Donation Button' to 'Donate here'. The field should accept free text and links to external websites.

- [ ] On the Project page this new 'Donate here' field should be shown in the Summary tab and in the Finances tab. The link to the external website should be clickable. It should only be shown if the user has input an external link in this field.

- [ ] The Finance tab on the Project page still has the 'Donations' field. It previously showed an accumulated amount of all donations done through this page. This feature should be disabled.

## Pending work

All the above mentioned items are completed, but a couple of other items that need to be done are below:

- ~~Clean up unused translation entries, and create new translation for Donate url~~.


punchagan:
Note: This PR is based on top of #2372 to avoid migration files causing conflicts.  #2372 should be easier to merge than this one, but merging this first should work okay too.  That PR can then be ignored.

punchagan:
I think the translations bit can be ignored for now.  There is some work required to be done to update the current state of translations in RSR to what our translators have changed.  Doing all the translations related work in a separate PR would be easier.

zzgvh:
Two small issues. 

First, it might be a good idea to "explain" the donation button that is now a link:
screenshot_137
When you click the donation button you're sent off to somewhere without explanation, and while I hate alerts that pop up when you're leaving a site in some places, maybe it's a good idea with an explanatory text? Any opinion @nadiagorchakova ?

Secondly, we have a few widgets. There is one with a donation button, but that seems to be missing now. 
rsr.akvo.org:
screenshot_139
rsr.localdev.akvo.org:
screenshot_138

punchagan:
1. An explanation might be useful, I agree. 
2. Thanks for the screenshots.  I'll look into this, and fix it.

punchagan:
@zzgvh The donate button missing in widgets should now be fixed. Thanks!

nadiagorchakova:
@zzgvh @punchagan having a popup after clicking on 'Donate' is a good idea. We should redirect to an external platform only if user clicks Ok on the popup.

punchagan:
I added a pop-up that looks like this:

mon oct 17 17 23 05 ist 2016

I'm not sure if the dialog text is good enough. Please suggest changes as you see fit, @nadiagorchakova @zzgvh 

Also, the popup doesn't appear after clicking on widgets since the widgets are anyway used in external sites and one would expect to be taken to an external website when clicking on them.

nadiagorchakova:
@punchagan Looks good. I would inform about the external platform first and then ask for action (donate/cancel). Seems to be more logical to me, otherwise a user might get confused why she has to click on Donate yet again. Would look like that:
'You will be sent to an external website that accepts donations for this project. To continue, click Donate'.

punchagan:
Changed the text as suggested.

zzgvh:
Two small issues remain. The modal doesn't appear when you click the Donate button on the listing page. The modal stays visible when you click Donate in it.

punchagan:
@zzgvh These two issues should be fixed now.

Issue #2375, opened 2016-09-20
Results framework performance issues


zzgvh:
This is a quick look into why the Results tab of a project in some cases is dog slow. Since I don't fully understand how Results are intended to be used, how they're implemented, and am doing this quick'n'dirty, some of my findings and conclusions may well be wrong.

I think the core of the problem lies in the [ResultSerializer](https://github.com/akvo/akvo-rsr/blob/1fea1c74410150b43078baf4d8704ab432d51cad/akvo/rest/serializers/result.py#L20). The ResultsSerializer includes four extra fields, i.e. artificial fields that are not part of the Result model fields, but instead are computed in some way: **project_title**, **type_label**, **parent_project** and **child_projects**.

The **project_title** field gets its data from the Project.title field. However looking at the [ResultsViewSet.queryset()](https://github.com/akvo/akvo-rsr/blob/1fea1c74410150b43078baf4d8704ab432d51cad/akvo/rest/views/result.py#L16) reveals that the queryset does not call select_related() so I think that a DB call is made for each Result to get the title of its associated project.

The **type_label** field calls [Result.iati_type()](https://github.com/akvo/akvo-rsr/blob/1fea1c74410150b43078baf4d8704ab432d51cad/akvo/rsr/models/result.py#L125) which is a wrapper that ultimately makes a Model.object.get() call on the relevant code list model, in this case ResultType.

The **parent_project** field calls [Result.parent_project()](https://github.com/akvo/akvo-rsr/blob/1fea1c74410150b43078baf4d8704ab432d51cad/akvo/rsr/models/result.py#L136) that checks if the FK field Result.parent_result holds any data and if so gets the ID and title of the project it points to. This might mean another DB call.

Finally the **child_projects** field calls [Result.child_projects()](https://github.com/akvo/akvo-rsr/blob/1fea1c74410150b43078baf4d8704ab432d51cad/akvo/rsr/models/result.py#L144). This innocent looking method makes one DB call to find all child *Result* objects, i.e. all Results that have this Result as parent. Then it proceeds to loop over this list of Result objects to find the ID and title of the Project linked to the Result. AFAICT this results in a number of DB calls equal to the number of children.

[...]

PR #2378, opened 2016-09-21
#2375 results performance

zzgvh:
Fixes the main issue with the **result** endpoint. There's still going to be lots of queries for results in complex parent-child project hierarchies, but at least they'll be fast. I think what happens is that the first Result queryset is cached and then used in all other queries.

### DIFF
--- a/akvo/rsr/models/result.py
+++ b/akvo/rsr/models/result.py
@@ -146,7 +146,7 @@ class Result(models.Model):
         Return a dictionary of this result's child projects.
         """
         projects = {}
-        for result in Result.objects.filter(parent_result=self).select_related('project__title'):
+        for result in Result.objects.filter(parent_result=self).select_related('project'):
             projects[result.project.id] = result.project.title
         return projects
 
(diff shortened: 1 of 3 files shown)


punchagan:
[Looks like](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#select-related) the related objects are fetched along with the first query.  :+1:

akvo/rsr/models/result.py
```diff
@@ -146,7 +146,7 @@ def child_projects(self):
         Return a dictionary of this result's child projects.
         """
         projects = {}
-        for result in Result.objects.filter(parent_result=self):
+        for result in Result.objects.filter(parent_result=self).select_related('project__title'):
```
punchagan: A nitpick: `select_related` is a way of asking related objects to be fetched in the same query, instead of making queries in the future. `select_related('project__title')` is therefore exactly equivalent to `select_related('project')` (verified by comparing the actual queries).  Can we drop the `__title`?

punchagan:
This great other than the one nit-picky comment I have!

akvo/rsr/models/result.py
zzgvh: Oh, interesting, I thought it'd only fetch the title field from project. But if that's not the case, then it's pointless of course.

PR #2383, opened 2016-09-28
Fix broken transaction import from IATI xml

punchagan:
If the DB doesn't have a row for the currency of a transaction already, the transaction can fail to be imported.


punchagan:
Closing in favor of #2395

Issue #2366, opened 2016-09-16
Arrows overlap in Project Hierarchy (investigate)


nadiagorchakova:
When Project Hierarchy is viewed from a child project page - arrows appear a bit messed up.
See http://rsr.akvo.org/en/project/4398/hierarchy/

![screen shot 2016-09-16 at 14 15 17](https://cloud.githubusercontent.com/assets/21127166/18585597/2c92a3d0-7c18-11e6-89e8-425afecde491.png)

nadiagorchakova:
- [ ] investigation is required

punchagan:
We have #2392 merged as a stop-gap fix for this.  What do we do with this issue?

nadiagorchakova:
The code with some small changes to arrows view was already deployed as part of the previous release. No need to investigate it further for now.

PR #2392, opened 2016-10-03
[#2366] Make arrows smaller and move them around

punchagan:
To avoid overlaps, the arrows are placed at slightly different positions
based on the index of the connector for which we are using an arrow, and
the curviness of the connectors is also changed.

| Before        | After         |
| ------------- |---------------|
| mon oct 3 16 58 30 ist 2016      | mon oct 3 16 57 39 ist 2016  |

### DIFF
--- a/akvo/rsr/static/scripts-src/project-hierarchy.js
+++ b/akvo/rsr/static/scripts-src/project-hierarchy.js
@@ -2,22 +2,28 @@ jsPlumb.ready(function() {
     var instance = jsPlumb.getInstance({
         Connector : [ "Bezier", { curviness:20 } ],
         DragOptions : { cursor: "pointer", zIndex:200 },
-        PaintStyle : { strokeStyle:"grey", lineWidth:2 },
-        EndpointStyle : { radius:3, fillStyle:"grey" },
+        PaintStyle : { strokeStyle:"grey", lineWidth:1 },
+        EndpointStyle : { radius:2, fillStyle:"grey" },
         HoverPaintStyle : {strokeStyle:"orange" },
         EndpointHoverStyle : {fillStyle:"orange" },
         Container:"project-hierarchy"
     });
 
     instance.doWhileSuspended(function() {
-        var arrowCommon = { foldback:0.7, fillStyle:"grey", width:7 },
+        var arrowCommon = { foldback:0.7, fillStyle:"grey", width:7, length:7 },
             arrow = [
                 [ "Arrow", { location:0.5 }, arrowCommon ]
-            ],
-            two_arrows = [
-                [ "Arrow", { location:0.7 }, arrowCommon ],
-                [ "Arrow", { location:0.3, direction:-1 }, arrowCommon ]
             ];
+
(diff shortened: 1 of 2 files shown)


punchagan:
This doesn't completely fix the problems with #2366, but mitigates them.  We could also add different colors for the connectors+arrows to make it clearer.  

But, to have curves which are not overlapping at all would be a lot more work.

akvo/rsr/static/scripts-src/project-hierarchy.js
```diff
@@ -2,23 +2,29 @@ jsPlumb.ready(function() {
...
-            two_arrows = [
-                [ "Arrow", { location:0.7 }, arrowCommon ],
-                [ "Arrow", { location:0.3, direction:-1 }, arrowCommon ]
             ];
 
+        // Function to off-set arrows based on the index of connector on the
+        // element for which we are using the arrows
+        var get_two_arrows = function (index) {
+            var two_arrows = [
+                [ "Arrow", { location:(0.9 - index*0.1) }, arrowCommon ],
```
punchagan: min(0.5, v) needed here

punchagan:
tue oct 4 12 55 08 ist 2016

This is one of the outputs I get by using entirely random colors.  We could be more methodical in choosing colors to make it look better, I guess. 

@loicsans might be able to give us more insights into this.

loicsans:
@punchagan, as far as I can see we have 2 size arrows.
One better solution would be to choose one colour for the short arrow and an other one for the long one.

punchagan:
tue oct 4 15 32 35 ist 2016

@loicsans We can have arrows of more than 2 sizes, actually, depending on how many siblings/children a project has.  But, I've implemented code that arrows of similar "size" are of the same color. ["size" is in quotes because, the library doesn't really let us detect size - the code indirectly interprets this, based on the index of the project in a list of sibling/child projects]

The colors I chose currently are random.  Since RSR doesn't have a color palette, etc. I'm not sure what colors to choose.  Any recommendations more than welcome!

zzgvh:
When I try this on project 3094, which has a ton of children, some of the lines only show when you hover over them:
rsr arrows
I'm seeing this in FF and Safari.

nadiagorchakova:
In the end, we decided not to implement different colours as it adds unnecessary complexity. Hierarchy for project 3094 looks funny no matter the colours though.

punchagan:
@zzgvh Thanks for trying this on that project.  Can you tell me how to import that data locally? 
It's because of the `curviness` being set to crazy numbers for the Bezier curves, I think. I'll look into this, tomorrow.

punchagan:
This is what we can get in the "small" case. 
thu oct 6 16 55 20 ist 2016

And this is what we get for the "large" case. 
thu oct 6 16 49 57 ist 2016

Getting anything better than this will be significantly more work, either with this or another library. (estimate: somewhere between 16 and 32). 

If the current results seem better than the older ones (which they do, to me) we can merge it, otherwise we can just close this PR.

PR #2394, opened 2016-10-06
[#2040] Add a js check for image upload size

### LINKED ISSUE #2040: Show error message if image file is too big when placing update (estimate: 6)
Currently, placing an update fails silently when the image attached is too big. We should add a nice error message for that.

punchagan:
This PR also adds a couple of new strings to be translated, but it doesn't do anything with them, because it would be easier to deal with all the translation issues in a separate issue.  

Closes #2040 

thu oct 6 13 28 35 ist 2016

thu oct 6 13 29 09 ist 2016

### DIFF
--- a/akvo/rsr/forms.py
+++ b/akvo/rsr/forms.py
@@ -289,10 +289,13 @@ class ProjectUpdateForm(forms.ModelForm):
         'placeholder': _(u'Description'),
         }))
     language = forms.ChoiceField(choices=settings.LANGUAGES, initial='en')
-    photo = forms.ImageField(required=False, widget=forms.FileInput(attrs={
-        'class': 'input',
-        'size': '15',
-        }))
+    photo = forms.ImageField(required=False,
+                             widget=forms.FileInput(attrs={
+                                 'class': 'input',
+                                 'size': '15',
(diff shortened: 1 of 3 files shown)


zzgvh:
I think the updated shouldn't be POSTable unless you remove an image that's too large. Makes it explicit that you have to fix it either way.

punchagan:
> I think the updated shouldn't be POSTable unless you remove an image that's too large. Makes it explicit that you have to fix it either way.

What exactly do you mean by the update not being POSTable? The Photo field isn't mandatory for the form, right now.  So, I'm not sure trying to upload a larger size photo should change that. 

The current code "unchooses" any file chosen by the user if it's larger than 2MB and changes the help text.  The UX isn't something that users normally see.  So, I am open to suggestions on how to do this better.  Ideally, the file dialog should only show images that are within the size limit, but I couldn't find a way to do this.

zzgvh:
Ah, I mean that the user has to remove the offending file from the form before it can be posted. That way it is obvious that the file is not part of the post. The error message can tell that the image has to be either smaller or removed.

punchagan:
Done.

Issue #2401, opened 2016-10-10
'NULL' in the last update column (akvoapp page)


nadiagorchakova:
Project 'Akvo PPP Phase 3 2014-2017' on [this](http://iggwater.akvoapp.org/en/projects/) website shows zero updates, whereas it has a number of updates linked to it. 

Lynn investigated the issue and it looks like all the projects shown under 'IGG' are showing 'NULL' in the last update column in the table 'rsr_project'. Only 'Akvo PPP Phase 3 2014-2017' project has updates so far and that's how the bug was detected.

<img width="1237" alt="screen shot 2016-10-10 at 10 35 31" src="https://cloud.githubusercontent.com/assets/21127166/19232510/6ad03b46-8ee0-11e6-8f21-31c13824a730.png">

zzgvh:
This is part of 3.17

PR #2402, opened 2016-10-11
#2401 fix latest update

zzgvh:
When the latest update for a project is deleted, the Project.last_update field gets set to None. This PR fixes the issue, adds a management command to set Project.last_update correctly and adds a test for the fix.

# Test plan

Select a project with updates. Note the latest and next to latest updates. Check that the latest update is linked in the project list. Delete the latest update in MyRSR. Check that the next to latest update is now the latest.

Also included is a management command, **check_latest_update** , that fixes any projects where the last_update field is None even if there are updates for the project. To test this, run the command. There will be an output of one character per project, where fixed projects are marked as **+**. Running this command twice there should be no **+** in the output the second time.


akvo/rsr/tests/models/test_project.py
```diff
@@ -0,0 +1,45 @@
...
+
+from akvo.rsr.models import Project
+from akvo.rsr.models import ProjectUpdate, PayPalGateway, MollieGateway, PaymentGatewaySelector
+
+
+class ProjectModelTestCase(TestCase):
+    """Tests for the project model"""
+
+    def test_project_last_update(self):
+        """ Test Project.last_update. The field is a denormalization keeping track of the latest
```
punchagan: When running the tests in the verbose mode, the first line of the doc string is printed out.  It would be nice to have a summary line, followed by an explanation ([as described here](https://www.python.org/dev/peps/pep-0008/#documentation-strings)).

akvo/rsr/management/commands/check_latest_update.py
```diff
@@ -0,0 +1,39 @@
...
+from django.core.management.base import BaseCommand
+from django.db.models import Prefetch
+
+from ...models import Project, ProjectUpdate
+
+class Command(BaseCommand):
+    help = "Set Project.last_update field. Fixes potential errors due to a bug"
+
+    def handle(self, *args, **options):
+        sys.stdout.write(u"Fix latest update for projects"
```
punchagan: I would call this command `fix_latest_update`, as the doc-string describes. :)

akvo/rsr/models/project.py
```diff
@@ -1469,3 +1469,16 @@ def update_denormalized_project(sender, **kwargs):
     project = project_update.project
     project.last_update = project_update
     project.save(last_updated=True)
+
+
+@receiver(post_delete, sender=ProjectUpdate)
+def rewind_last_update(sender, **kwargs):
+    """ Updates the denormalized project.last_update on related project,
```
punchagan: As with the test doc-string, there may be other tools in the eco-system which assume the first line is a summary line.

akvo/rsr/tests/models/test_project.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
```
punchagan: `akvo/rsr/tests/models/` will also need an `__init__.py`

punchagan:
The changes look good to me, but I don't know why these changes caused the build to fail the way it did. :confused:  
Restarted a build to see if it passes.

PR #2428, opened 2016-11-07
#2424 Make queryset filtering based on permissions faster

### LINKED ISSUE #2424: Checking for permissions of private objects is super slow
When a user is logged in and not an rsr_admin (or superuser), all the viewsets for them go bonkers when there are a significant number of private projects.  
A bazillion queries are made to the DB to identify whether or not an object should be included in a queryset by checking if the user has permissions to change it.  

See https://github.com/akvo/akvo-rsr/blob/develop/akvo/rest/viewsets.py#L214

punchagan:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

When `private_objects` were being filtered out of querysets of ViewSets subclassed from `PublicProjectViewSet`, every `private_object` in the queryset was iterated over and the permissions for the current user for it were verified.  As the number of private projects increases, this will get slower and slower.  

This commit tries to convert this Python loop into DB queries, to make the filtering of the `queryset` faster. The permission rules are parsed and converted to a `queryset` filter using `Q` objects and the filtering of objects for which a user doesn't have permissions happens within DB queries.  To ensure nothing has changed w.r.t permissions from before, tests have been added for `queryset`s of all `ViewSet`s subclassing from `PublicProjectViewSet`. 

Closes #2424

### DIFF
--- a/akvo/rsr/models/user.py
+++ b/akvo/rsr/models/user.py
@@ -4,20 +4,22 @@
 # See more details in the license.txt file located at the root folder of the Akvo RSR module.
 # For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
 
+import re
+
 from django.contrib.auth.models import AbstractBaseUser, BaseUserManager, PermissionsMixin, Group
 from django.core.mail import send_mail
 from django.db import models
 from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
-
+import rules
 from sorl.thumbnail.fields import ImageField
-
 from tastypie.models import ApiKey
 
 from akvo.utils import rsr_image_path
 
 from .employment import Employment
 from .project_update import ProjectUpdate
+from .project import Project
 
 from ..fields import ValidXMLCharField, ValidXMLTextField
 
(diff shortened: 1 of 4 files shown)


akvo/rsr/models/user.py
```diff
@@ -4,20 +4,22 @@
 # See more details in the license.txt file located at the root folder of the Akvo RSR module.
 # For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
 
+import re
+
 from django.contrib.auth.models import AbstractBaseUser, BaseUserManager, PermissionsMixin, Group
 from django.core.mail import send_mail
 from django.db import models
```
zzgvh: I would import the Q object explicitly from django.db.models since it's used in so may places.

akvo/rsr/tests/rest/test_permissions.py
```diff
@@ -0,0 +1,450 @@
...
+from akvo.utils import check_auth_groups
+
+
+def slugify(s):
+    return re.sub('\W+', '-', s.lower())
+
+
+User = get_user_model()
+
+
```
zzgvh: This is just because I don't like anonymous tuples 😉... 

The counts tuples could be made into namedtuples:
```python
import collections
# gc is the number of objects visible to that group, groupcount
gc = collections.namedtuple('gc', 'admin anonymous editor other')
...
            model_map[M.Project] = {'counts': gc(8, 4, 6, 4), 'project_relation': ''}
```
etc.

Then iter_queryset() can be nicely shortened:

```python
def iter_queryset(self, user_type='other'):
    """An iterator over querysets to test permissions with."""

    for model in self.model_map:
        data = self.model_map[model]
        count = getattr(data['counts'], user_type)
        yield model.objects.all(), data['project_relation'], count
```

akvo/rsr/tests/rest/test_permissions.py
```diff
@@ -0,0 +1,450 @@
...
+                                 user.username, queryset.model.__name__))
+
+    def test_logged_in_user(self):
+        from akvo.rest.viewsets import PublicProjectViewSet
+        m_e = Group.objects.get(name='M&E Managers')
+        p_e = Group.objects.get(name='Project Editors')
+        admins = Group.objects.get(name='Admins')
+        for user in User.objects.filter(is_admin=False, is_superuser=False):
+            # If not project editor, then counts are same as anonymous user
+            user_type = 'editor' if user.in_group(m_e) or user.in_group(p_e) or user.in_group(admins) else ''
```
zzgvh: Must change to
```python
user_type = 'editor' if user.in_group(m_e) or user.in_group(p_e) or user.in_group(admins) else 'other'
```
to work with the namedtuples fix.

zzgvh:
Line lengths are a bit longish in places too. I try to stay under 100 chars...

punchagan:
@zzgvh addressed all your comments. The namedtuples fix is good! Thanks for suggesting it.

PR #2437, opened 2016-11-10
[#2398] Fix users not showing when a user has no groups

### LINKED ISSUE #2398: No users shown in user management page
When an employment doesn't have a group (See #2122, #2374), the User management page in MyRSR doesn't display any of the users.  Some js assumes that all employments have a group.

punchagan:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

The user management page doesn't show any users when a single employment
shown on a page doesn't have a group assigned.  The post_save signal on
Employments actually tries to ensure that all Employments actually have
a Group and uses the 'Users' group as default.  But, the current DB
seems to have at least one Employment without a Group.

Fixes #2398 

Ideally, this should come with a data migration to fix any employments that don't have a group assigned. But, I'm not sure if we can should get that in, before a release.  I'll try working on that, now. This bug seemed like something that's an edge case that won't happen in the live DB, but it is currently happening. Hence, a "quick-fix".

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,6 +8,10 @@
 number of projects in data quality reports if an organisation has multiple
 partnership roles for a project.
 
+[#2398] (https://github.com/akvo/akvo-rsr/issues/2398) Fix empty user
+management page when a user in a page has an employment that does not have a
+group defined.
+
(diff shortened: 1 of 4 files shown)


punchagan:
I've implemented the migration, as well.

Issue #2442, opened 2016-11-15
Custom import of IATI results


zzgvh:
The EUTF want to be able to import Result data _only_ using the IATI import. The regular import mappers will delete most content of a project if only ```<result>``` tags are in the XML.

To solve this a custom mapper is needed along with some modifications to IatiImport.model_mappers()

punchagan:
Closed with #2445

PR #2443, opened 2016-11-15
#2442 only mapper

zzgvh:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
I want to write a test case for this, but it'd be good if you could merge this @punchagan, if it looks OK, so we can get it on test asap.

PR #2458, opened 2016-11-24
[#2381] Use only approved organisation as legacy_org

### LINKED ISSUE #2381: Legacy API user organisation may be unapproved (estimate: 3)
#2364 fixed the problem where Up version 1.5 shows random organisations for Update authors. However, the org id supplied may be for an organisation for which the user is not (yet) approved, which is potentially confusing. It would be be better if the legacy org was chosen only from those approved.

punchagan:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

Closes #2381


zzgvh:
A changelog entry maybe? Don't think we need a test for it.

punchagan:
Done.

punchagan:
Merging myself, since the change is 1) tiny 2) already reviewed by @zzgvh.  The only change requested was a release notes entry, which I added.

Issue #2487, opened 2016-12-16
Invalid HTTP_HOST header: '*.live.akvo-ops.org'. 


punchagan:
The domain name provided is not valid according to RFC 1034/1035.

punchagan:
Closed with #2488

PR #2488, opened 2016-12-16
[#2487] Fix problem with disallowed hosts in requests

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

Move request.get_host into the try block where it was originally meant to be.

### DIFF
--- a/akvo/rsr/tests/test_middleware.py
+++ b/akvo/rsr/tests/test_middleware.py
@@ -46,24 +46,23 @@ class HostHeaderTestCase(TestCase):
 
     """Testing boot traffic."""
 
-    def setUp(self):
-        """Setup."""
-        self.c = Client(HTTP_HOST='_')
-
     def test_underscore_host(self):
         """When host is '_'."""
+        self.c = Client(HTTP_HOST='_')
         resp = self.c.get('/')
-        self.assertEqual(resp.status_code, 400)
+        self.assertEqual(resp.status_code, 302)
 
     def test_empy_host(self):
         """When host is ''."""
-        resp = self.c.get('')
-        self.assertEqual(resp.status_code, 400)
+        self.c = Client(HTTP_HOST='')
+        resp = self.c.get('/')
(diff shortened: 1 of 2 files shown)


zzgvh:
[These are the tests that fail](https://github.com/akvo/akvo-rsr/blob/%232487-fix-invalid-hosts/akvo/rsr/tests/test_middleware.py#L53). But do we really want to get rid of them? test_underscore_host() looks wrong also, should be ```resp = self.c.get('_')``` shouldn't it?

punchagan:
The tests were actually failing because some code in the TestClient's handler was failing for invalid request headers.  In the actual server code, we are catching the exception DisallowedHost and redirecting the user to the default RSR domain.  But the test client code fails when it tries to create a response object because it does things looking at the request too. (I didn't look into the details of why exactly the request needs to be looked at). 

I pushed a fix that changes the request objects headers before doing a redirect, to fix the tests.  This shouldn't hopefully break anything for normal users.

akvo/rsr/tests/test_middleware.py
```diff
@@ -46,24 +46,23 @@ class HostHeaderTestCase(TestCase):
...
+        resp = self.c.get('/')
+        self.assertEqual(resp.status_code, 302)
 
     def test_oddchar_host(self):
         """When host is ''."""
-        resp = self.c.get('$')
-        self.assertEqual(resp.status_code, 400)
+        self.c = Client(HTTP_HOST='$')
+        resp = self.c.get('/')
+        self.assertEqual(resp.status_code, 302)
```
zzgvh: Would it make sense to add this test:

```python

    def test_wildcard_host(self):
        """When host is ''*.live.akvo-ops.org"""
        self.c = Client(HTTP_HOST='*.live.akvo-ops.org')
        resp = self.c.get('/')
        self.assertEqual(resp.status_code, 302)
```

just to spell out exactly the problem we had?

PR #2495, opened 2016-12-28
[#2494] Switch to DB queries to get allow_edit_projects

### LINKED ISSUE #2494: Signing in from RSR Up fails with error 'Unable to authorize:'
Login for a user who has `rsr.change_project` permissions for a lot of projects, but is not an `rsr_admin` can timeout. This piece of code seems like the [culprit code](https://github.com/akvo/akvo-rsr/blob/5a07b31d1c272f2014e5eb3209f23506aa6e91f1/akvo/rsr/views/account.py#L273).

punchagan:
To get the list of projects where a user has `change` permissions, use DB
queries instead of a loop wrapping DB queries to make the response much faster.
This fixes timeouts for users who are not rsr admins but have change
permissions for a lot of projects.

Closes #2494

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/tests/views/test_account.py
+++ b/akvo/rsr/tests/views/test_account.py
@@ -1,0 +1,142 @@
+# -*- coding: utf-8 -*-
+
+"""
+Akvo RSR is covered by the GNU Affero General Public License.
+
+See more details in the license.txt file located at the root folder of the Akvo RSR module.
+For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
+"""
+
+from __future__ import print_function
+
+import json
+
+from django.conf import settings
+from django.contrib.auth.models import Group
+from django.test import TestCase, Client
+
+from akvo.rsr.models import Employment, Organisation, Partnership, Project, User
+from akvo.utils import check_auth_groups
+
+
+class AccountTestCase(TestCase):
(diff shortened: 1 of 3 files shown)


akvo/rsr/tests/views/test_account.py
```diff
@@ -0,0 +1,141 @@
...
+        self.org1 = Organisation.objects.create(name='akvo', long_name='akvo foundation')
+        self.org2 = Organisation.objects.create(name='icco', long_name='icco foundation')
+        for org in (self.org1, self.org2):
+            project = Project.objects.create()
+            project.publishingstatus.status = 'published'
+            project.publishingstatus.save()
+            Partnership.objects.create(project=project, organisation=org)
+        self.c = Client(HTTP_HOST=settings.RSR_DOMAIN)
+
+    def tearDown(self):
```
zzgvh: Nitpick: should the Groups be deleted here too?

punchagan: `check_auth_groups` actually checks if groups exist before trying to create them. So, it doesn't really matter, but I guess for consistency, it's better to delete them. I'll push a fix.

punchagan: Done.

PR #2509, opened 2017-01-19
[#2507] Fix broken pagination on subclasses of PublicProjectViewSet

### LINKED ISSUE #2507: Indicator period not shown in RSR Results page
One partner reported that indicator period is not showing up on RSR results page.
Please see project - http://rsr.akvo.org/en/myrsr/project_editor/5285/

Each indicator in their Results framework seems to have indicator period 01/01/2016 - 31/12/2016 but it does not appear in the Results tab and consequently their akvoapp page: http://nso-g4aw.akvoapp.org/en/myrsr/results/5285/#results,20918,16601 
All other indicator periods appear as normal. 
So far, I wasn't able to detect what's different about year 2016 for this project..

@punchagan, could you investigate?

punchagan:
Not ensuring that `.distinct` is called on the queryset used by subclasses of
`PublicProjectViewSet` caused a bug where an indicator period was missing from
the results page of a project.  This commit fixes that bug.

This commit comes with a simple test, but the test isn't actually able to
reproduce the bug on production.  Not sure what exactly needs to be done to
reproduce that bug.

Closes #2507

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rest/viewsets.py
+++ b/akvo/rest/viewsets.py
@@ -139,13 +139,13 @@ class PublicProjectViewSet(BaseRSRViewSet):
         if user.is_anonymous() or not (user.is_superuser or user.is_admin):
             queryset = self.projects_filter_for_non_privileged_users(user, queryset, self.project_relation)
 
-        return queryset
+        return queryset.distinct()
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


punchagan:
Closed in #2523

PR #2515, opened 2017-01-20
[#2514] Return unicode in RecipientCountry.iati_country_unicode

### LINKED ISSUE #2514: UnicodeDecodeError at /rest/v1/recipient_country/
'ascii' codec can't decode byte 0xc3 in position 7: ordinal not in range(128)

punchagan:
Closes #2514

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/models/country.py
+++ b/akvo/rsr/models/country.py
@@ -51,7 +51,7 @@ class RecipientCountry(models.Model):
         'Project', verbose_name=_(u'project'), related_name='recipient_countries'
     )
     country = ValidXMLCharField(
-        _(u'recipient country'), blank=True, max_length=2,choices=codelist_choices(COUNTRY, show_code=False),
+        _(u'recipient country'), blank=True, max_length=2, choices=codelist_choices(COUNTRY, show_code=False),
         help_text=_(u'The country that benefits from the project.')
     )
     percentage = models.DecimalField(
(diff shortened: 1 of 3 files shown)


akvo/rsr/models/country.py
```diff
@@ -85,7 +85,7 @@ def iati_country(self):
         return codelist_value(codelist_models.Country, self, 'country')
 
     def iati_country_unicode(self):
-        return str(self.iati_country())
+        return unicode(self.iati_country())
```
zzgvh: Do we really need unicode() here? After all self.iati_country() returns the codelists.models.Country.name field. Should be usable as is, no?

akvo/rsr/tests/rest/test_recipient_country.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
```
zzgvh: Should we really test the API here? It's the RecipientCountry.iati_country() method that's the root problem.

punchagan: @zzgvh - as discussed on Skype, leaving this test as is, since it tests both things.

akvo/rsr/models/country.py
punchagan: @zzgvh `self.iati_country()` is actually calling `codelist_value`, not `codelist_name`, and therefore returns and instance of `codelists.models.Country`.  The `unicode` call is required.

zzgvh: Ah, right!

PR #2522, opened 2017-01-25
[#2338] Add report types and options to the database

### LINKED ISSUE #2338: Add report types and options to database
Currently, the report types are hard coded in a python file. However, it should be possible to add / edit / delete these reports without needing a release. Therefore, these report types (and other options) should be added to a new model and stored in the database.

punchagan:
Closes #2338

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry


akvo/rsr/models/report.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+from ..fields import ValidXMLCharField, ValidXMLTextField
+
+
+class ReportFormat(models.Model):
+    name = ValidXMLCharField(_('name'), max_length=30, unique=True)
```
zzgvh: This is kind of a nit-pick, but I think this field should either be named **format**, since that's how it's used or **document_type** because that's what it is.

akvo/rsr/static/scripts-src/my-reports.jsx
```diff
@@ -180,7 +180,7 @@ function initReact() {
                     );
                 } else {
                     return (
-                        <span key={format.key} />
```
zzgvh: :bug: (I can't put the comment where the bug actually is...) Above on line 176 format.displayName should be format.display_name.

akvo/rsr/models/report.py
```diff
@@ -0,0 +1,42 @@
...
+    title = ValidXMLCharField(_(u'title'), max_length=200)
+    description = ValidXMLTextField(
+        _(u'description'), blank=True, help_text=_(u'Describe the report.')
+    )
+    url = ValidXMLCharField(
+        _(u'url'),
+        max_length=200,
+        help_text=_(u'Enter the parametrized path for downloading the report'),
+    )
+    parameters = ValidXMLCharField(
```
zzgvh: It would be really cool if the format parameter was added automatically since it's mandatory (I think?). Either that or check for its presence I think would be good. Less work is to just update the help text to remind that format must be present 😉

akvo/rsr/models/report.py
punchagan: `ReportFormat.name` or `format.name` seemed better to me than `format.format` or `ReportFormat.format`.  I'm leaning towards leaving `name` as it is. It may just be me, and if you insist, I can change this.

akvo/rsr/static/scripts-src/my-reports.jsx
punchagan: Good catch. Thanks!

akvo/rsr/models/report.py
punchagan: I'm actually going to just remove the `parameters` field, and make it a property so that the `serializer` can return this data, but the user needn't bother about it. `url` field is going to have validation to ensure that the parameters are "valid".

akvo/rsr/models/report.py
zzgvh: Hehe, no it's fine. I think document_type is probably the best, but let's leave it for now!

akvo/rsr/models/report.py
zzgvh: Ah, good solution.

PR #2541, opened 2017-02-17
[#2540] Allow listing more projects in the directory

### LINKED ISSUE #2540: Projects directory listing only shows 10 projects
The project listing directory is hard coded to show only 10 projects.  This limit seems to be in place due to some changes made in #1258 and may not be necessary any more.

punchagan:
The query for `partnership_more_link` is very slow, and adding a
`select_related` makes it quite a lot faster.  This looks like the only query
that is slow on the project listing page.

We add a `limit` parameter to specify the number of projects to be listed in
the directory, with the default value being 10.

Closes #2540

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


akvo/rsr/tests/views/test_project.py
```diff
@@ -0,0 +1,27 @@
...
+
+from __future__ import print_function
+
+from unittest import skip
+
+import django_perf_rec
+from django.test import TestCase
+
+
+@skip('Needs Django >= 1.8')
```
zzgvh: Cool! We need to upgrade!

akvo/rsr/views/project.py
```diff
@@ -96,7 +96,8 @@ def directory(request):
 
     # Build page
     page = request.GET.get('page')
-    page, paginator, page_range = pagination(page, sorted_projects, 10)
+    limit = request.GET.get('limit', 10)
```
zzgvh: Maybe have the default value come from settings?

akvo/rsr/models/project.py
```diff
@@ -981,13 +981,9 @@ def budget_total(self):
...
 
     def has_multiple_budget_currencies(self):
-        budget_items = BudgetItem.objects.filter(project__id=self.pk)
-        num_currencies = len(set([self.currency] + [c.currency if c.currency else self.currency for c in budget_items]))
-
-        if num_currencies > 1:
-            return True
-        else:
-            return False
+        budget_items = self.budget_items.all()
```
punchagan: It's a trade off between making an extra query vs. using the `budget_items` which were already pre-fetched.  We probably need to measure which one is better for the average case. I'll play around a little, with this code, and get back.

akvo/rsr/views/project.py
punchagan: :+1:

akvo/rsr/models/project.py
punchagan: Looks like the Python iteration is about a hundred times faster, when the `budget_items` have been pre-fetched and there are only a couple of them.

zzgvh: Right and we're getting them anyway. Interesting!

PR #2553, opened 2017-02-21
[#1853] Show only organisation projects on Akvo sites search

### LINKED ISSUE #1853: Search for projects in Akvo Pages
At the Aqua for All Page http://aquaforall.akvoapp.org/en/projects/ there are currently 209 projects displayed. 

In the search bar I can search for ANY project we've got in the RSR database. For example I can search for an Akvo project in which Aqua for All is not a partner. 
screen shot 2015-10-09 at 12 05 54

That means I can see any project within the Aqua for All Akvo Page, e.g. http://aquaforall.akvoapp.org/en/project/2849/ 

This should not be the case. 

I would expect I can only search for projects and organisations that are part of the 209 projects.

### LINKED ISSUE #2526: Search in EUTF akvo site
Partner team had a training and workshop with EUTF last week and discovered that search terms in EUTF akvo site returned unrelated results.

Search for tombouctou shows up a project of SNV in EUTF akvo page, which is confusing for the partner as they expect to see their own projects only on their akvo site. 

screen shot 2017-02-06 at 15 56 41

What the partner expects is to see just projects where they are one of the participating partners. 
If the search does not match any of their projects, it should then not return anything.

punchagan:
Fixes #1853
Fixes #2526

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rest/views/typeahead.py
+++ b/akvo/rest/views/typeahead.py
@@ -13,6 +13,7 @@ from akvo.rest.serializers import (TypeaheadCountrySerializer,
 
 from akvo.codelists.models import Country, Version
 from akvo.rsr.models import Organisation, Project, ProjectUpdate
+from akvo.rsr.views.project import _project_directory_coll
 
 from django.conf import settings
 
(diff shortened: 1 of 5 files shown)


akvo/rest/views/typeahead.py
```diff
@@ -59,7 +59,15 @@ def typeahead_user_organisations(request):
 
 @api_view(['GET'])
 def typeahead_project(request):
-    projects = Project.objects.all().exclude(title='')
+    from akvo.rsr.views.project import _project_directory_coll
+    if request.GET.get('published', '0') == '0':
```
zzgvh: Shouldn't the projects be filtered by employment and group in the project editor? So that you don't get info on projects you can't access anyway. This becomes important if we expand the scope of the data returned by the typeahead I think.

zzgvh: Come to think of it this is also relevant for EUTF where one of the complaints is that they get projects that aren't "theirs" when searching in MyRSR. At first I thought we might leave it with the fixes you did, but I think we need to be stricter in the project editor too.

PR #2579, opened 2017-03-17
Add buttons to view more projects in the directory

### LINKED ISSUE #2577: Add links to be able to view more projects in the project directory
Currently, only 10 projects are shown in the project directory. Add an option to be able to view more projects.

punchagan:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

Closes #2577


zzgvh:
I'm a little concerned about this:

screenshot_159

This might take forever on a slow connection...

punchagan:
I'm not sure what the best way of going about this would be. Setting chrome to limit the speed to GPRS and trying with just 10 projects is also pretty slow!  It would be unbearable with 100 projects, I agree.

Also, the UX/UI/CSS can use some work from @loicsans or @Kiarii.

punchagan:
image

Added a screenshot of the current UI.

zzgvh:
The UI is fine as far as I'm concerned. Maybe the current number of projects is shown as text, then the list of alternatives remains the same except for the current choice not being a link.

zzgvh:
As for the partners I think we should investigate this data being part of the initial response. The current version seems very wasteful when you have many projects with the same partners.

punchagan:
I tried implementing something where the data about partners is sent in the template.  I tried to keep the changes minimal, as of now.

akvo/templates/project_directory.html
```diff
@@ -264,6 +271,26 @@ <h4 class="detailedInfo">{% trans "Finance" %}</h4>
         }
     </script>
 
+    {# More partners data #}
+    {% for project in page %}
+    <script type="application/json" id="more-partners-{{project.id}}">
```
zzgvh: The whole script tag should be copied into project_main.html since the partner tooltip is used there too!

akvo/templates/project_directory.html
```diff
@@ -264,6 +271,26 @@ <h4 class="detailedInfo">{% trans "Finance" %}</h4>
...
+    [
+    {% for partnership in project.partnerships.distinct %}
```
zzgvh: .distinct seems to destroy the prefetch here and I can't see that it's needed. I replaced it with .all and got an order of magnitude faster response. 

If you think it should be distinct it has to be part of the query using the [Prefetch](https://docs.djangoproject.com/en/1.7/ref/models/queries/#django.db.models.Prefetch) class I think.

akvo/templates/project_directory.html
```diff
@@ -78,6 +78,13 @@
       {% blocktrans with start_index=page.start_index end_index=page.end_index paginator_count=paginator.count %}
           Viewing {{ start_index }} - {{ end_index }} of {{ paginator_count }} projects
       {% endblocktrans %}
+          | View
+          {% with "20 50 100" as page_sizes %}
```
zzgvh: Two issues here. 

I think the settings.PROJECT_DIRECTORY_DEFAULT_SIZE page size should be included in the list.

When changing page size your "cursor" is eliminated. By this I mean that the page counter is kept the same but the page size changes so you unless you're on the first page you will see a totally different set of projects. Ideally you'd have the same first project when you switch page size but I can see that might be hard to do given the rather blunt page=X&limit=Y, but maybe it's worth trying to calculate the page so that the first project on the old page is part of the new page.

akvo/templates/project_directory.html
punchagan: You are right about `.distinct` destroying the prefetch, and it looks like we don't really need a distinct here. Good catch! Thanks.

akvo/templates/project_directory.html
punchagan: :+1:

akvo/templates/project_directory.html
punchagan: I see what you are saying, and it makes a lot of sense. I'll try to get this done.

Issue #2405, opened 2016-10-13
Standardize on one country codelist


zzgvh:
We currently have two lists of countries: [ISO_3166_COUNTRIES](https://github.com/akvo/akvo-rsr/blob/806930ac46b02f033c8c83d711334adc2dc41145/akvo/rsr/iso3166.py#L9) that is a hand crafted list based on the ISO 3166-1 alpha-2 list, and [COUNTRY](https://github.com/akvo/akvo-rsr/blob/806930ac46b02f033c8c83d711334adc2dc41145/akvo/codelists/store/codelists_v202.py#L89) that is part of the IATI codelists.

We should only use one codelist and I suggest standardizing on the IATI codelist.

janagombitova:
Created via Reamaze:

Link: https://flowhelp.reamaze.com/admin/conversations/enhancement-geolocation-question-type

PR #2589, opened 2017-03-24
Add Country Field to Locations and automatically populate it

### LINKED ISSUE #2466: Add 'Recipient Country' field to the API (estimate: 16)
ICCO has reported that their project country data does not get through the API anymore. The 'missing' country field seems to be a legacy issue as it was removed from Editor some time ago, while the API was not updated. 

We should add 'Recipient country' from section 7 in the Editor to the API code. It should also be marked that the old country field must not be used.

(It might be that the original country field was removed as part of this issue https://github.com/akvo/akvo-rsr/pull/2277)

### LINKED ISSUE #2527: Filter by country in EUTF akvo site (estimate: 16)
EUTF would like to be able to filter by country using Advanced filters on their akvo site (http://eutf.akvoapp.org).

Currently, the only option in the location filter is 'world':

screen shot 2017-02-06 at 16 23 56

I checked akvo sites for other partners like Washalliance, and the country filter works there, see: 

screen shot 2017-02-06 at 16 25 57

Should be on EUTF akvo site as well. @Lynngre could you please have a look too?

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

This pull request adds code to automatically figure out the country for a `Location` (or subclass) instance using it's `latitude` and `longitude`.  

This pull request may still need some work, specially w.r.t the following things
- We use a dataset from okfn to figure out the country boundaries.  How accurate is this, isn't entirely clear. Randomly testing some locations, against Google maps, seemed to work. What would be a good test for this? 
- What point in the code is a good place to download (and read) this dataset? 
- How often should this file get updated? 
- Do we need to display the automatically computed country to the user? Or can it be hidden in the DB? 

Closes #2466 
Closes #2527


punchagan:
> The one glaring remaining sillyness I wonder if we can get rid of is that there are now two country fields for Organisations! (The same is technically true for Projects if you include recipient countries, but let's leave that for now, it kinda works...)

I agree this is pretty silly, and I did try removing it. But, I think this is a part of the bigger issue of having two country code lists.  OrganisationLocation already had two locations before this change was made, though one of them was deprecated (using a comment saying so :wink: ) 

> We also have two code lists for countries.
> Thoughts?

This is indeed a problem, but fixing this seems like a separate issue to me. I'd say we go ahead and merge this, since it fixes #2527 for EUTF and #2466 (for some other partners).  I'll make a new issue for cleaning up code lists for countries, if we don't already have one.

punchagan:
Seems like we already have an issue for this - #2405

zzgvh:
Yep, I agree, let's get this into develop!

PR #2591, opened 2017-03-29
Issue/2548 keyword search filter

zzgvh:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rsr/static/scripts-src/project-directory.js
+++ b/akvo/rsr/static/scripts-src/project-directory.js
@@ -33,29 +33,6 @@ $(document).ready(function () {
     var organisations_text = templateJSON.organisations_text;
     var locations_text = templateJSON.locations_text;
 
-    var locations = new Bloodhound({
-        datumTokenizer: Bloodhound.tokenizers.obj.whitespace('name'),
-        queryTokenizer: Bloodhound.tokenizers.whitespace,
-        prefetch: {
-            //  TODO: get a filtered set of locations for Pages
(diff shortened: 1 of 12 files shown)


akvo/rsr/static/scripts-src/project-directory.js
```diff
@@ -3,190 +3,148 @@
...
+
+            filter: function (response) {
+                return response.results;
+            }
+        }
+    });
+
+    projects.initialize();
+    locations.initialize();
+    keywords.initialize();
```
punchagan: Are we leaving out the organisations deliberately? Or were they missed out accidentally? 

Like I mentioned on Skype, I think, we should only have typeaheads for the project in the project directory search bar.  That will make it clear that the search is only looking in the project titles, and the typeahead is a way of making this search faster (by doing it on client side).  Similarly for the organisation and the project update directory.  The other typeahead fields are already in the advanced filter and should be used from there. 

This change, though, may need more discussion, especially with @loicsans and @AkvoAnt. Until, then I think we should keep the UI as before with the keywords typeahead added to it.

zzgvh: Yes, I dropped the organisations. But I can easily put them back. One issue with that is that the search results popup become so tall it might drop below the lower edge of the browser 😞  We can mitigate that by only showing 3 items of of each, but that has other issues, obviously...

PR #2601, opened 2017-03-31
[#2356] Enable publish & add project buttons only if user has perms

### LINKED ISSUE #2356: Publish button
The Publish button in Project Editor should be shown as inactive when user doesn't have rights to publish a project.
As a Project Editor at Akvo, I see the seemingly clickable Publish button, however, when I click on it nothing happens.
As a Project Editor at a partner organisation, e.g. 1% club, I do not see the Publish button at all.

screen shot 2016-11-16 at 11 14 53

When partner team members in the hubs do trainings, they sometimes report the "missing" Publish button as a bug.

### LINKED ISSUE #2362: Add a new project button shown even if no permissions
When a user (is an admin, but) doesn't belong to any organisation that `can_create_projects`, the `Add a new project` button shouldn't be shown. 

Currently, when that is clicked, the project editor doesn't open up and the button gets de-activated. 

But, when the projects listing page is reloaded a new empty project is shown. 

thu sep 15 18 59 55 ist 2016

This may not be as much of a problem on the deployed instance, but was something I found on the dev instance.

punchagan:
- [ ] Copyright header 
- [ ] Code formatting 
- [ ] Documentation 
- [ ] Change log entry

Closes #2356 
Closes #2362

### DIFF
--- a/akvo/rsr/models/user.py
+++ b/akvo/rsr/models/user.py
@@ -376,16 +376,11 @@ class User(AbstractBaseUser, PermissionsMixin):
         return self.employers.all().exclude(is_approved=False)
 
     def can_create_project(self):
-        """
-        Check to see if the user can create a project, or is a superuser.
-
-        :return: Boolean to indicate whether the user has a reportable organisation
-        """
-        if self.is_superuser or self.is_admin:
-            return True
+        """Check to see if the user can create a project."""
 
         for employment in self.approved_employments():
-            if employment.organisation.can_create_projects:
+            org = employment.organisation
+            if org.can_create_projects and self.has_perm('rsr.add_project', org):
                 return True
 
         return False
--- a/akvo/rsr/models/user.py
+++ b/akvo/rsr/models/user.py
@@ -378,9 +378,6 @@ class User(AbstractBaseUser, PermissionsMixin):
     def can_create_project(self):
         """Check to see if the user can create a project."""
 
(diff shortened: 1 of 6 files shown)


akvo/rsr/models/user.py
```diff
@@ -376,16 +376,14 @@ def approved_employments(self):
...
 
     def can_create_project(self):
-        """
-        Check to see if the user can create a project, or is a superuser.
+        """Check to see if the user can create a project."""
 
-        :return: Boolean to indicate whether the user has a reportable organisation
-        """
-        if self.is_superuser or self.is_admin:
+        if self.is_superuser:
```
zzgvh: Let's keep god-mode. Not sure about self.is_admin, I think this is Akvo staff only and if so it might apply to them too?

punchagan: `self.has_perm` automatically gives all permissions to any user who is a `superuser`.  Also, if a user `is_admin` they have the `rsr.add_project` permissions given to them in our permission config.  So, I don't think this special case is required.

punchagan: When adding a new project, approved employments of a user are used to determine which organisation to add the project to. Having an admin/super user without any employments trying to add a new project will be a problem, if we have this special case.

zzgvh: Ah...OK you have to have an employment then I guess.

PR #2623, opened 2017-04-06
Fix problems with Location advanced filter in project directory

punchagan:
`ProjectLocation`s associated with projects were being used to populate the list of options in the advanced filter for locations, while `RecipientCountry` locations are shown in the project directory, when showing information about each project.  This made the search buggy, and also hard to understand.  To make it easier for the user to understand how search is working, this code ignores the `ProjectLocation` countries completely, and simply uses the `RecipientCountry` values.

- [ ] Copyright header 
- [ ] Code formatting 
- [ ] Documentation 
- [ ] Change log entry

Closes #2623

### DIFF
--- a/akvo/rsr/filters.py
+++ b/akvo/rsr/filters.py
@@ -9,12 +9,14 @@ see < http://www.gnu.org/licenses/agpl.html >.
 from copy import deepcopy
 
 import django_filters
+from django.db.models import Q
 from django.utils.translation import ugettext_lazy as _
 
 from akvo.codelists.store.codelists_v202 import ACTIVITY_STATUS, SECTOR_CATEGORY
 from akvo.utils import codelist_choices
 from .models import (Category, Keyword, Organisation, OrganisationLocation,
-                     Project, ProjectUpdate, ProjectUpdateLocation, RecipientCountry)
+                     Project, ProjectLocation, ProjectUpdate, ProjectUpdateLocation,
+                     RecipientCountry)
 from .m49 import M49_CODES, M49_HIERARCHY
 
 ANY_CHOICE = (('', _('All')), )
@@ -61,7 +63,12 @@ def filter_m49(queryset, value):
     if not value:
         return queryset
     countries = walk(deepcopy(M49_HIERARCHY)[int(value)])
-    return queryset.filter(recipient_countries__country__in=countries)
+    countries_lower = [c.lower() for c in countries]
+    filter_ = (
+        Q(recipient_countries__country__in=countries) |
+        Q(locations__country__iso_code__in=countries_lower)
+    )
+    return queryset.filter(filter_)
 
 
 def filter_m49_orgs(queryset, value):
@@ -98,10 +105,9 @@ def get_locations(location, locations):
 def location_choices(qs):
     """From a queryset get possible location filter choices"""
 
-    country_ids = (
-        get_recipient_country_ids(qs) if qs.model is Project else
-        get_location_country_ids(qs)
-    )
+    country_ids = get_location_country_ids(qs)
+    if qs.model is Project:
+        country_ids += get_recipient_country_ids(qs)
 
     locations = [
         location
@@ -124,7 +130,10 @@ def get_recipient_country_ids(projects):
 def get_location_country_ids(qs):
     """Return countries for locations associated with objects in the queryset."""
 
-    if qs.model is ProjectUpdate:
+    if qs.model is Project:
+        location_model = ProjectLocation
+
+    elif qs.model is ProjectUpdate:
         location_model = ProjectUpdateLocation
 
     elif qs.model is Organisation:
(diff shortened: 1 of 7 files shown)


akvo/rsr/filters.py
```diff
@@ -98,26 +98,45 @@ def get_locations(location, locations):
 def location_choices(qs):
     """From a queryset get possible location filter choices"""
 
-    if qs.model is Project:
-        location_model = ProjectLocation
-    elif qs.model is ProjectUpdate:
+    country_ids = (
+        get_recipient_country_ids(qs) if qs.model is Project else
```
zzgvh: I think we should concatenate the country IDs from recipient country and project_location. Otherwise we exclude around 100 projects from being found when filtering on country.

```sql
select p.id from rsr_project p, rsr_publishingstatus s where p.id not in (select DISTINCT p.id from rsr_project p, rsr_recipientcountry r where p.id = r.project_id) and p.id = s.project_id and s.status = 'published'
```

punchagan: That was my initial solution, but I removed the project locations because they are not shown in the project description in the project directory.  Do you think we should show them?  

image

zzgvh: Ya when there's no recipient country I think we should. It's a bit clunky but otherwise a number of projects will be "locationless". (I haven't checked how many projects have no country associated at all...)

zzgvh: I ran this query (WARNING it's a bomb, takes almost three minutes to run the query on my vagrant!):

```sql
SELECT p.id
FROM rsr_project p, rsr_publishingstatus s
WHERE p.id NOT IN (SELECT DISTINCT p.id
                   FROM rsr_project p, rsr_recipientcountry r, rsr_projectlocation l
                   WHERE p.id = r.project_id OR p.id = l.location_target_id) AND p.id = s.project_id
      AND s.status = 'published'
```

 on a month old RSR DB but that's had the country-from-coords migration applied. 41 published projects have no country at all 😢

punchagan: I think it's alright that filtering by location doesn't show all the projects. For instance, this might happen with keywords or sector filters too -- not all projects may have an associated keyword/sector.

To keep the filter simple, I ignored the Project Locations, but I think we can append that to the list of recipient countries. If we are searching on the Project Locations, too, I think they should always be shown in the list of locations for a project in it's listing in the project directory.

punchagan: If we change the list of locations shown in the project listing in the directory, should we also change the list of locations in the project page's header? 

image

zzgvh: Sure the filter doesn't show all project but I think that if a project is associated with a country it should be possible to find. And by extension I'd say that if there's no recipient country but we have a country from project location we should display that.

akvo/rsr/filters.py
```diff
@@ -98,26 +105,47 @@ def get_locations(location, locations):
 def location_choices(qs):
```
zzgvh: This function is very hard to understand in general. I think it's part of the old code base, right?

akvo/rsr/filters.py
```diff
@@ -98,26 +105,47 @@ def get_locations(location, locations):
...
     """From a queryset get possible location filter choices"""
 
+    country_ids = get_location_country_ids(qs)
+    if qs.model is Project:
+        country_ids += get_recipient_country_ids(qs)
+
```
zzgvh: I think applying list(set(country_ids)) here should shorten the list

akvo/rsr/filters.py
```diff
@@ -98,26 +105,47 @@ def get_locations(location, locations):
...
+    locations = [
```
zzgvh: This part I don't even understand, but doint locations = list(set(locations)) at line 117 should also shorten the list

akvo/rsr/filters.py
```diff
@@ -98,26 +105,47 @@ def get_locations(location, locations):
...
-        if location.country:
-            country = get_id_for_iso(location.country.iso_code.upper())
-            locations.append(country)
-            locations.extend(get_locations(country, []))
-
-    choices = [tup for tup in M49_CODES if any(
-        unicode(i) in tup for i in locations)]
-    return [M49_CODES[0]] + choices  # Add the world to the choices
+    return [
+        get_id_for_iso(location.country.iso_code.upper())
```
zzgvh: This blows up if location has no country

akvo/rsr/models/project.py
```diff
@@ -644,6 +644,18 @@ def set_reporting_org(self, organisation):
...
 
+    def countries(self):
+        """Return a list of countries for the project."""
+
+        country_codes = set([c.country.lower() for c in self.recipient_countries.all()])
+        return (
+            [country for country in self.recipient_countries.all()] +
+            [
+                location.country for location in self.locations.all()
+                if location.country.iso_code not in country_codes
```
zzgvh: This can also blow up

akvo/rsr/filters.py
punchagan: This has a check for location.country in the if clause, and shouldn't blow up.

akvo/rsr/filters.py
punchagan: I've tried to make the code a little more readable. Let me know if there's anything else that can be done to make it more so.

PR #2647, opened 2017-04-27
Do not list/count private projects anywhere outside of MyRSR

punchagan:
- Private projects are meant to be viewed only by members of the organisations which are partners of the projects. Members of these organisations could be shown private projects in the project listings, etc., but to keep these pages simple, it was decided to exclude private projects from all the pages outside of MyRSR.

- The existence of private projects shouldn't be known to anyone outside of the organisations to which they "belong".  So, all public counts of projects must exclude private projects.

- [x] Copyright header

- [x] Code formatting

- [x] Documentation

- [x] Change log entry

### DIFF
--- a/akvo/rest/views/typeahead.py
+++ b/akvo/rest/views/typeahead.py
@@ -96,7 +96,7 @@ def typeahead_project(request):
     if request.GET.get('published', '0') == '0':
         # Project editor - organization projects, all
         page = request.rsr_page
-        projects = page.organisation.all_projects() if page else Project.objects.all()
+        projects = page.all_projects() if page else Project.objects.all()
     else:
         # Search bar - organization projects, published
         projects = _project_directory_coll(request)
--- a/akvo/rsr/models/organisation.py
+++ b/akvo/rsr/models/organisation.py
@@ -497,35 +497,9 @@ class Organisation(TimestampsMixin, models.Model):
             amount_pledged=Sum('partnerships__funding_amount')
         )['amount_pledged'] or 0
 
-    def euros_pledged(self):
-        "How much € the organisation has pledged to projects it is a partner to"
-        return self.active_projects().euros().filter(
-            partnerships__organisation__exact=self,
-            partnerships__iati_organisation_role__exact=Partnership.IATI_FUNDING_PARTNER
-        ).aggregate(
-            euros_pledged=Sum('partnerships__funding_amount')
-        )['euros_pledged'] or 0
-
-    def dollars_pledged(self):
-        "How much $ the organisation has pledged to projects"
-        return self.active_projects().dollars().filter(
-            partnerships__organisation__exact=self,
-            partnerships__iati_organisation_role__exact=Partnership.IATI_FUNDING_PARTNER
-        ).aggregate(
-            dollars_pledged=Sum('partnerships__funding_amount')
-        )['dollars_pledged'] or 0
-
     def org_currency_projects_count(self):
         "How many projects with budget in default currency the organisation is a partner to"
         return self.published_projects().filter(currency=self.currency).distinct().count()
-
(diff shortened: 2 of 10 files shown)


akvo/rest/views/typeahead.py
```diff
@@ -96,7 +96,7 @@ def typeahead_project(request):
     if request.GET.get('published', '0') == '0':
         # Project editor - organization projects, all
         page = request.rsr_page
-        projects = page.organisation.all_projects() if page else Project.objects.all()
+        projects = page.all_projects() if page else Project.objects.all()
```
zzgvh: Are we still OK with private projects being available here?

akvo/rsr/models/organisation.py
```diff
@@ -278,13 +277,7 @@ def knowledge(self):
...
             "returns a queryset with all projects that has self as any kind of partner"
             from .project import Project
-            return Project.objects.filter(partnerships__organisation__in=self).distinct()
-
-        def public_projects(self):
-            """
-            Returns a queryset with all public projects that has self as any kind of partner.
-            """
-            return self.all_projects().public()
+            return Project.objects.of_partner(self).distinct()
```
zzgvh: Question on the underlying code: Since Project.of_partner() actually is Project.QuerySet.of_partner(), it should work with a queryset of organisations as param, right? But if so, isn't it identical in function to Project.of_partners()?

akvo/rsr/models/partner_site.py
```diff
@@ -181,17 +181,21 @@ def partners(self):
...
+    def all_projects(self):
         """All projects of the Page"""
         from .project import Project
         # Get all projects associated via the Page's organisation
         if self.partner_projects:
-            fk_projects = self.organisation.published_projects().public()
+            fk_projects = self.organisation.all_projects()
         else:
-            fk_projects = Project.objects.public().published()
+            fk_projects = Project.objects.all()
```
zzgvh: Doesn't this result in private projects being included if self.partner_projects is false? I guess it comes down to where all_projects() is used.

akvo/rest/views/typeahead.py
punchagan: If we are showing unpublished projects also, I suspect private projects are fine. 
May be it doesn't make sense to add unpublished/private projects as related projects (that's where this end-point seems to be used).

akvo/rsr/models/partner_site.py
punchagan: This is related to the comment above, and that is where this is being used.

akvo/rsr/models/organisation.py
punchagan: We have a function `of_partners` and `of_partner` on `Project.QuerySet`.  `of_partners` works with a list/queryset of organisations, while `of_partner` returns all the projects of a single organisation. Methods on the `QuerySet` essentially need to return a `queryset` of the same model, while methods on the model itself can do literally anything.  I don't exactly understand what the question is, but I don't see a problem with having this method.

zzgvh: Ya, but what I mean is that since of_partner can take a queryset of many orgs as param doesn't it do the same thing as of_partners?

punchagan: Ah, sorry. I get the question, now! Yes, they are essentially doing almost the same thing. Except, passing a single organisation to `of_partners` won't work, since it expects an iterable. 

```python
org = Organisation.objects.get(id=1)
Project.objects.all().of_partners(org)  # will fail
Project.objects.all().of_partners([org]) # works
Project.objects.all().of_partner(org) # same as above
```

akvo/rsr/models/partner_site.py
zzgvh: Right, makes sense.

PR #2649, opened 2017-04-28
Improve web page speed using CDN and Cache-Control

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


punchagan:
image

There's still some things that we can't enforce caching for, because it's served off of other domains/servers. But, this commit significantly improves on the caching.

punchagan:
The CDN usage could probably be improved to some extent, and I'll try and see how much I can get done before we freeze for the release.

punchagan:
Code in lots of pages expects `#react-typeahead` element to be available. I pushed a commit moving out react-typeahead.js from the `rsr_lib` aggregated script. 
In the (near) future, we may decide to drop this library in favor of `react-bootstrap-typeahead.js` which has a CDN version too. (See #2650)

PR #2650, opened 2017-05-03
QuickFix to make the My results filter less "sensitive"

punchagan:

- [x] Copyright header

- [x] Code formatting

- [x] Documentation

- [x] Change log entry


punchagan:
The first commit is a simple fix to use substring matches instead of fuzzy search, using the same library, to make the search less sensitive. 

I also pushed a second commit, which introduces the use of a new typeahead library - `react-bootstrap-typeahead.js`.  The `react-typeahead.js` library is used in multiple places in RSR and all those uses should be replaced with this one, if we go ahead and merge this commit.  This library seems better than the older one - more similar to the twitter typeahead library (jquery based) which we use for the project/organisation/update directory typeaheads.

PR #2673, opened 2017-05-18
Admin action to set projects complete

zzgvh:
- [x] Copyright header
- [x] Code formatting
- [ ] Documentation
- [x] Change log entry

```markdown
Admin action to mark projects as complete [#2671](https://github.com/akvo/akvo-rsr/issues/2671)
```

### DIFF
--- a/akvo/rsr/admin.py
+++ b/akvo/rsr/admin.py
@@ -33,6 +33,7 @@ from sorl.thumbnail.fields import ImageField
 from embed_video.admin import AdminVideoMixin
 import os.path
 
+from admin_actions import set_project_status_complete
 from akvo.rsr.mixins import TimestampsAdminDisplayMixin
 from akvo.utils import custom_get_or_create_country
 from akvo.rsr.fields import ValidXMLCharField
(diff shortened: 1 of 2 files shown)


punchagan:
@zzgvh I tested this feature and it looks okay to me.  Can you add a note to add to the RELEASE_NOTES ?

Also, 
1. Why do we want to set the status of projects to complete? Is there a specific reason to do this? From the RSR code, it doesn't look like completed projects don't get much special treatment. 

2. Is it okay to expose this feature without having an undo? If a project is accidentally set to complete, I don't see an easy way of going back. The `iati_status` field seems to override the `status` field in the old admin interface for the project and the project editor.

zzgvh:
1) The original reason is a request from Annabelle. There are 100+ projects for the WASH alliance and they want all of them set to complete.

2) Since it's an internal tool, to be used only by Akvo staff in the old admin, I'd say it's OK not to have an undo. But sure, you can mess things up if you're not careful.

3) The [save method](https://github.com/akvo/akvo-rsr/blob/develop/akvo/rsr/models/project.py#L457) on Project fixes this, and I thought that was actually a good idea, making sure that both status fields get updated. 

However, when you mention it, it's probably better to do it the other way around, setting the iati_status explicitly and status implicitly in save() since iati_status is what we expose. I think status is kept for legacy reasons that may not even be around anymore...

zzgvh:
Realized Project.save() isn't called when using queryset.update() 😒

punchagan:
Looks good to me. Merging.

Issue #2651, opened 2017-05-03
User management page times out when a user has many employments


punchagan:
The permissions system and the user_management view both make a huge number of queries when a user has multiple employments.  This causes the request to timeout. 

Created via Reamaze:

Link: https://akvoo.reamaze.com/admin/conversations/editor-rights

Message:
Dear Akvo team,

Over the last couple of days/weeks I tried several times to give a colleague editor rights for AKVO RSR. Apart from the system being very slow, I don’t succeed in opening the ‘user management’ tab, or, in the rare cases I succeeded, I was not able to give editor rights.  In all these cases I got the message  ‘504 gateway time-out’.

Thank you very much
Best regards,

PR #2685, opened 2017-05-22
[#2651] Don't cache groups in the code

punchagan:
Groups can be deleted and recreated -- this happens in tests, though this may
not happen in the real code.  This breaks tests because cached group ids are not
the same as the current ids of groups in the db.

Instead, check for permissions using group names.  Groups are fetched from the
DB using names, everywhere.  We therefore move to comparing employment group
name with canonical names for checking permissions.

Fixes the broken tests on develop. 

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/permissions.py
+++ b/akvo/rsr/permissions.py
@@ -7,17 +7,15 @@
 import rules
 
 from django.contrib.auth import get_user_model
-from django.contrib.auth.models import Group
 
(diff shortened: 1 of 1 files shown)


punchagan:
@zzgvh This PR fixes the broken tests on develop.  Updated the code to compare group names, instead of using cached group ids when checking for permissions.

PR #2692, opened 2017-05-26
[#2666] Remove actual value and comment fields from project editor

### LINKED ISSUE #2666: Disable 'Actual Value' and 'Actual Value Comment' fields from Project Editor (estimate: 8)
actual value fields

punchagan:
We could've set some hidden fields in the validation sets, but since we are
changing the template to prevent the page from looking weird, we are going to
need to change back the template if we are going to add those fields back. The
change cannot be made without a deploy, and we might as well change the template
in that case.

Closes #2666

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry
```markdown
Remove the actual value and actual value comments from the project editor [#2666](https://github.com/akvo/akvo-rsr/issues/2666)
```

### DIFF
--- a/akvo/templates/myrsr/project_editor/related_objects/indicator_period_input.html
+++ b/akvo/templates/myrsr/project_editor/related_objects/indicator_period_input.html
@@ -24,23 +24,7 @@
                 {% include text_input with obj=period field='target_value' %}
             </div>
             <div class="col-md-6">
-                {% if not period.data.exists %}
-                    {% include text_input with obj=period field='actual_value' %}
-                {% else %}
-                    {% include text_input with obj=period field='actual_value' disabled=True %}
-                    <p class="help-block">
-                        {% trans 'It is not possible to edit the actual value when indicator data has been filled in through the My results section. ' %}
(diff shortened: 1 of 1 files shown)


punchagan:
*NOTE*: as per discussion with @AkvoAnt, this should be merged only when the results framework is out of Beta.

zzgvh:
Code reviewed. Looks good. Won't merge now.

punchagan:
This would mean that we are moving away from setting an actual value directly on a period.  We should probably have a data migration for this, and change all periods which have such data, and instead use an update to save this data.  This would simplify the aggregation code (both on the backend and the frontend), since we will only have to aggregate update values then.

zzgvh:
Yep, sounds good!

zzgvh:
There are other consideration here though. Particularly in relation to the old Results framework page. I'd say we should stop using it before we make these kinds of structural changes, otherwise we have to do work on the old results that will very soon be "thrown away".

AkvoAnt:
Migrate actual values and comments in periods entered in the project editor to indicator updates.

PR #2698, opened 2017-05-30
[#2697] Show organisation specific reports only to those org users

### LINKED ISSUE #2697: Show organisation specific reports only to users of those organisations
There's already a Plan Finland report, which is currently filtered out using a UI hack.  #2505 and #2655 are going to add an EUTF specific report. We need a better way of displaying a organisations specific report options only to the users who are employed by the organisations.

punchagan:
This commit adds an organisation field to reports, to associate one or more
organisations with a report. The report option is displayed only to users of
those organisations or to admin users.

Closes #2697

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/models/report.py
+++ b/akvo/rsr/models/report.py
@@ -39,6 +39,7 @@ class Report(models.Model):
         help_text=_(u'Enter the parametrized path for downloading the report'),
     )
     formats = models.ManyToManyField(ReportFormat)
+    organisations = models.ManyToManyField('Organisation', blank=True)
 
     @property
     def parameters(self):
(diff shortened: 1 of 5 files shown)


akvo/rsr/models/report.py
```diff
@@ -39,6 +39,7 @@ class Report(models.Model):
         help_text=_(u'Enter the parametrized path for downloading the report'),
     )
     formats = models.ManyToManyField(ReportFormat)
+    organisations = models.ManyToManyField('Organisation')
```
zzgvh: This field needs to be null=True, blank=True otherwise it's not possible to save an instance without any org

punchagan: @zzgvh I don't think a ManyToManyField needs `null=True` or `blank=True`. I couldn't find a reference in the official documentation, but I found this [SO answer](https://stackoverflow.com/a/18244527/1225068), and it makes sense to me.

punchagan: I guess `blank=True` is required, though, for the admin interface to work

PR #2720, opened 2017-06-06
[#2719] Enable submitting the test env password modal using Return

### LINKED ISSUE #2719: Make the password dialog on test environments more usable
It looks like the password protection on the Test Environments was added as [a warning to users](https://github.com/akvo/akvo-rsr/blob/develop/RELEASE_NOTES.md#password-protection-for-test-environments) that the data being viewed is not real data, but test data.  In that case, I'm not sure a password is required. We could have an alert like the cookie usage warning that we show. 
The modal dialog doesn't stop any bots from actually crawling the sites, either. 

Before we decide to replace it with something easier to use, we can enable auto focus, and submit on return, to make it slightly easier to deal with.

punchagan:
Also, automatically focus the password input when the dialog is shown.

Closes #2719

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry
```markdown
Make the password dialog on test environments more usable [#2719](https://github.com/akvo/akvo-rsr/issues/2719)
```

### DIFF
--- a/akvo/rsr/static/scripts-src/cookie.js
+++ b/akvo/rsr/static/scripts-src/cookie.js
@@ -114,9 +114,10 @@ function createModal(){
                         React.createElement('br'),
                         React.createElement('br'),
                         React.createElement(
-                            'div',
+                            'form',
                             {
-                                className: formGroupClass
+                                className: formGroupClass,
+                                onSubmit: this.checkPassword
                             },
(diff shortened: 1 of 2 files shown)


punchagan:
Seems like, I did. Fixed that now! Thanks for catching it, @zzgvh

PR #2727, opened 2017-06-15
[#2713] Add a progress bar to the indicator header

### LINKED ISSUE #2713: Actual vs. Target progress bar (0 to 100%) for the 'Overview'
screen shot 2017-06-02 at 15 26 56

punchagan:
Closes #2713

Initial PR for adding a progress bar. 

The view currently looks like this. 
image

    - View the results page of a project with indicators which have target values filled in for periods
   - Progress bar should be shown for all indicators which have at least one period with a target value
   - No progress bar is shown for indicators with no target value periods
   - 100% or more completion should show a green progress bar 
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry
```
Add a progress bar using the actual and target values of an indicator - [#2713](https://github.com/akvo/akvo-rsr/issues/2713)
```


punchagan:
image

Circular progress indicators

punchagan:
- [ ] Equal length
- [ ] Vertically aligned
- [ ] Subtler colors
- [ ] In the same line as the title

punchagan:
image

akvo/rsr/static/scripts-src/my-results/selectors.js
```diff
@@ -133,7 +133,47 @@ export const getIndicatorsAggregateActualValue = createSelector(
...
+
+export const getIndicatorsAggregateTargetValue = createSelector(
+    /*
+      Return an object on the form:
+      {indicatorId1: <aggregateTargetValue1>, indicatorId2: <aggregateTargetValue2>,...}
+    */
+    [getIndicatorIds, getIndicatorsChildrenIds, getPeriodObjects],
+    (indicatorIDs, childPeriodIds, periodObjects) => {
+        return indicatorIDs.reduce((acc, indicatorId) => {
+            const aggregateValue = childPeriodIds[indicatorId].reduce((acc, periodId) => {
```
zzgvh: This _might perhaps_ blow up if childPeriodIds is ever undefined. The problem is that I'm not sure. If you look further down at **getNeedReportingPeriods** for example I do this
```javacript
    ...
    unlockedPeriods && updateObjects && unlockedPeriods.filter(
    ...
```
to make sure things don't go south if the params are undefined. In the case of **getNeedReportingPeriods** this could happen if you opened the tree before all models were loaded (I think, it might have been some of the other selectors, but you get the drift). How to handle uninitialized objects and arrays is a general headache that I don't have a good solution for. It's probably something we should try to solve though.

akvo/rsr/static/scripts-src/my-results/components/Indicators.jsx
```diff
@@ -29,34 +32,41 @@ import {
...
-            </li>
-        </ul>
-    :
-        undefined;
+                        <ul>
+                            <li className="description">
+                                <span>{indicator.description}</span>
+                            </li>
+                        </ul>
+              :
```
zzgvh: Heh, this poor colon is kinda hangin' all by itself 😉 When I write multiline ternary statements I see everything up to and including ? as the if: and the : as the else: and indent as if it's python 😛

Issue #2730, opened 2017-06-15
Prevent interaction before fetching of all models is complete


zzgvh:
When loading the Results page there is a more or less short window of time where the UI is rendered but not fully functional. This may lead to the page breaking if the user interacts "too soon". 

To prevent this replace the accordion with a Loading label until model fetching from the back end is completed.

punchagan:
Closed in #2731

PR #2736, opened 2017-06-17
[#2730] Improve UI interaction before all models are loaded

zzgvh:
Add Loading... messages on all accordion levels and add more checks to selectors making sure they work when the params they rely on are not populated yet.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### Testing

This is a little hard to test. I used the throttling in Chrome to load the page at a crawl and then I got a short time window when not all models were loaded. As far as I could tell I could still interact with the page, opening the Results accordion and see how the aggregate values in the indicators appeared when the periods and updates were fetched.

Proper unit tests would improve the situation here 😉.

### DIFF
--- a/akvo/rsr/static/scripts-src/my-results/components/App.jsx
+++ b/akvo/rsr/static/scripts-src/my-results/components/App.jsx
@@ -311,10 +311,10 @@ export default class App extends React.Component {
             <div className="col-xs-6">
             </div>;
 
-        const results = this.props.ui.allFetched ?
-            <Results parentId="results"/>
-        :
-            <p className="loading">Loading <i className="fa fa-spin fa-spinner" /></p>;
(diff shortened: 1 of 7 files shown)


akvo/rsr/static/scripts-src/my-results/components/App.jsx
```diff
@@ -311,10 +311,10 @@ export default class App extends React.Component {
             <div className="col-xs-6">
             </div>;
 
-        const results = this.props.ui.allFetched ?
-            <Results parentId="results"/>
-        :
-            <p className="loading">Loading <i className="fa fa-spin fa-spinner" /></p>;
+        // const results = this.props.ui.allFetched ?
```
punchagan: This commented code can simply be deleted.

PR #2737, opened 2017-06-20
[#2684] Keep the set page size value even after filters applied

### LINKED ISSUE #2684: Active filter should be kept when changing number of projects viewed
When changing the number of projects shown in the project directory, any filter that is in force should be "brought along".

punchagan:
Page size is set using a `limit` query parameter. When filters are applied, a
form submission happens, and only the form data is submitted to the server. The
`limit` query parameter gets lost.  This commit adds a hidden field for the
limit parameter to the form, so that this data is retained even when form
submissions happen.

Closes #2684

  - Go to the project directory
  - Change the number of projects being view from 10 to 20/50/100. 
  - Select one or more advanced filters and apply them
  - Check that the results pages have the same number of projects per page, as chosen above

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry
```
Don't reset the page size to the default value of 10, when filters are applied [#2684](https://github.com/akvo/akvo-rsr/issues/2684)
```

### DIFF
--- a/akvo/rsr/views/project.py
+++ b/akvo/rsr/views/project.py
@@ -128,6 +128,7 @@ def directory(request):
         'current_org': org_filter,
         'map_projects': map_projects,
         'sectors_dict': sectors_dict,
+        'limit': limit,
     }
     return render(request, 'project_directory.html', context)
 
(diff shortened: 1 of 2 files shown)


punchagan:
Ah, I wasn't aware of this @iperdomo. Thanks!

PR #2743, opened 2017-06-22
[#2740] Rename query parameter iati_status to status

### LINKED ISSUE #2740: iati_status query parameter is still used on the site
The query parameter for status was changed from `iati_status` to `status` in #2681. But, the `iati_status` filter is being used in a bunch of places (for example on an organisation page - the link to projects). 

`iati_status` parameter needs to be brought back, and can be renamed to `status` when processing it.

punchagan:
The filter name was changed in a previous
commit (fa8094647b1270ec5aa28f15c09b10f7f1b24693) when the dynamic filter was
implemented.  But, a lot of links actually use the old filter, and hence are
broken.  This commit renames the query parameter to fix the problem.

Closes #2740

  - The links to projects in different states (completed/post-completion, etc.) on the Organisation page should show the correct number of projects. 
    For example: https://rsr.akvo.org/en/organisation/42

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry
```
Fix broken links to projects in different states from the organisation page [#2740](fa8094647b1270ec5aa28f15c09b10f7f1b24693)
```

### DIFF
--- a/akvo/rsr/views/project.py
+++ b/akvo/rsr/views/project.py
@@ -45,6 +45,10 @@ def directory(request):
     """The project list view."""
     qs = remove_empty_querydict_items(request.GET)
 
+    # iati_status was renamed to sector in fa8094647b
+    if 'status' not in qs:
+        qs['status'] = qs.pop('iati_status', [''])[0]
+
     # Set show_filters to "in" if any filter is selected
     show_filters = "in"  # To simplify template use bootstrap class
     available_filters = [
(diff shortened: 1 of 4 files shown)


akvo/rsr/views/project.py
```diff
@@ -45,6 +45,9 @@ def directory(request):
     """The project list view."""
     qs = remove_empty_querydict_items(request.GET)
 
+    # iati_status was renamed to sector in fa8094647b
+    qs['status'] = qs.pop('iati_status', [''])[0]
```
zzgvh: This overwrites status with '' if there is no iati_status param. This is the case when applying the advanced filter on the project listing page.

punchagan: Thanks for the catch, @zzgvh! Yes, I guess I wasn't thinking clearly when I made this change. Fixed this, now.

PR #2744, opened 2017-06-22
Use a cookie to detect if advanced filters pane should be open or not

punchagan:
  1. Click on the advanced search button to open the advanced filters pane. 
  2. Reload the page to see that the advanced filters pane is open.
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry
```
Open the advanced filters pane if it was previously opened, not otherwise [#2741](https://github.com/akvo/akvo-rsr/issues/2741)
```

### DIFF
--- a/akvo/rsr/static/scripts-src/rsr.js
+++ b/akvo/rsr/static/scripts-src/rsr.js
@@ -33,12 +33,27 @@ window.AKVO_RSR.utils = {
 
 $(document).ready(function() {
 
+  function toggle_advanced_filters(show){
+    // Toggles the advanced filters pane if show is undefined
+    // Shows the pane if show is true.
+    // Hides the pane if show is false.
+    if (show != undefined && show == $("#wrapper, #search").hasClass('toggled')){
+      return;
+    }
+    $("#wrapper, #search").toggleClass("toggled");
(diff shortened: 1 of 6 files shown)


akvo/rsr/static/scripts-src/rsr.js
```diff
@@ -38,6 +38,7 @@ $(document).ready(function() {
     e.preventDefault();
     $("#wrapper, #search").toggleClass("toggled");
     $("a.showFilters > i").toggleClass("fa-toggle-off fa-toggle-on");
+    Cookies.set(advancedFilterCookie, $("#wrapper, #search").hasClass('toggled')?'on':'off');
```
zzgvh: This splats, since advancedFilterCookie is undefined.

punchagan: Sorry for the broken PR. I think I had a local change, which I forgot to commit or something.

punchagan: This should be fixed now.

punchagan:
This was a bug in the code, that didn't handle pages which didn't have the pane/sidebar. I fixed this now, and should work as expected. Thanks for the extensive testing, Gabriel!

PR #2749, opened 2017-06-29
[#2738] Use actual value of a period directly

### LINKED ISSUE #2738: Results and Indicators Overview Report is not pulling the results data correctly from the New Results Framework
The results data appearing in the Results and Indicators Overview doesn't seem to be pulling the correct data saved in the new results framework: 

screen shot 2017-06-21 at 11 51 46

screen shot 2017-06-21 at 11 52 09

punchagan:
Instead of aggregating the updates on the client side, we directly use the
actual value on the period, so that

1. we remove duplication of this aggregation computation on the client side
2. handle the case when a period has no updates, but the actual value has been
   set directly on the period.

Closes #2738

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

```markdown
Use the actual value for a period entered in project editor if there is one and no updates have been added to the period. [2738](https://github.com/akvo/akvo-rsr/issues/2738)
```

### DIFF
--- a/akvo/rsr/static/scripts-src/my-results/selectors.js
+++ b/akvo/rsr/static/scripts-src/my-results/selectors.js
@@ -104,8 +104,8 @@ export const getPeriodsActualValue = createSelector(
         Return an object on the form:
         {periodId1: <actualValue1>, periodId2: <actualValue2>,...}
      */
-    [getPeriodIds, getUpdateObjects, getPeriodsChildrenIds],
-    (periodIds, updateObjects, childUpdateIds) => {
+    [getPeriodIds, getPeriodObjects, getUpdateObjects, getPeriodsChildrenIds],
+    (periodIds, periodObjects, updateObjects, childUpdateIds) => {
         return periodIds && updateObjects && !isEmpty(childUpdateIds) && periodIds.reduce((acc, periodId) => {
             const actualValue = childUpdateIds[periodId].filter(
                 (updateId) => updateObjects[updateId].status == c.UPDATE_STATUS_APPROVED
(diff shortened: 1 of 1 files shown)


akvo/rsr/static/scripts-src/my-results/selectors.js
```diff
@@ -100,29 +100,17 @@ export const getUpdatesChildrenIds = createSelector(
...
-                    if (!(data !== data)) {
-                        return acc + data;
-                    }
-                    return acc;
-                }, 0
-            );
-            return {...acc, [periodId]: actualValue};
-        }, {});
-    }
+  /*
```
zzgvh: The problem with this is that period.actual_value isn't updated when we add or change an update. So we'd have to reload the period from the API after posting updates if this is to work.

punchagan: Currently, I fixed it by just using the actual value of the period, when there are no updates (in any state).  
At some point, I wonder if it would be better to remove the aggregation completely from the client side code, to remove duplication. But, it won't be as fast/responsive as it is now.

akvo/rsr/static/scripts-src/my-results/selectors.js
```diff
@@ -120,7 +120,8 @@ export const getPeriodsActualValue = createSelector(
                     return acc;
                 }, 0
             );
-            return {...acc, [periodId]: actualValue};
+          const periodActualValue = (parseFloat(periodObjects[periodId].actual_value)||0);
```
zzgvh: A comment here describing what's going on would be good I think. It's kinda non obvious...

punchagan: Done!

PR #2773, opened 2017-07-19
[#2771] Display organization long names in the partners tab

### LINKED ISSUE #2771: Project listing should show long names where possible
Created via Reamaze:

Link: https://akvoo.reamaze.com/admin/conversations/project-partners-short-and-long-names
Assignee: Anthony Gonzalez

Message:
AfDB is one of our RSR partners. They load IATI files into RSR. When partner names are loaded, the organisation name is automatically abbreviated. An example is: 

https://rsr.akvo.org/en/organisation/3502/ 

The long name is also filled out. The partner suggested to display the long name instead of the short name on the partners tab by default. Example: 

https://rsr.akvo.org/en/project/5485/#partners 

An alternative approach which makes more sense perhaps is to show both the short and long names on the partners tab.

punchagan:
Closes #2771

  - Check that the name of the organisation Rural Water Supply & Sanitation is displayed fully in https://rsr.akvo.org/en/project/5485/#partners
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry
```
Display long names for organisations (whenever available) in the project partners tab -- [#2771](https://github.com/akvo/akvo-rsr/issues/2771)
```

### DIFF
--- a/akvo/rsr/models/organisation.py
+++ b/akvo/rsr/models/organisation.py
@@ -79,7 +79,7 @@ class Organisation(TimestampsMixin, models.Model):
     name = ValidXMLCharField(
         _(u'name'), max_length=40, db_index=True, unique=True,
         help_text=_(u'Short name which will appear in organisation and partner listings '
-                    u'(25 characters).')
+                    u'(40 characters).')
     )
     long_name = ValidXMLCharField(
         _(u'long name'), max_length=100, db_index=True, unique=True,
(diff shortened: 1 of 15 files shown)


akvo/rsr/static/scripts-src/project-main/project-main-partners.jsx
```diff
@@ -121,16 +124,18 @@ function renderPartnersTab() {
...
                     }
 
+                    var id = partner[0].organisation.id;
+
                     return (
-                        <div className='row verticalPadding projectPartners'>
+                        <div className='row verticalPadding projectPartners' key={id}>
                             <div className="col-sm-2 img">
-                                <a href={'/en/organisation/' + partner[0].organisation.id + '/'}>
+                                <a href={'/en/organisation/' + id + '/'}>
```
zzgvh: Nitpick: I think we should remove /en here. I know it currently doesn't matter, but one could conceive of a future where we show different logos depending n language choice. Same goes for the name below.

zzgvh:
I've implemented some "intelligence" to determine if we use Organisation.name or .long_name. The second commit implements this in the Organisation model and applies it in a number of places.

akvo/rsr/models/organisation.py
```diff
@@ -321,7 +321,31 @@ def content_owned_organisations(self):
             return queryset
 
     def __unicode__(self):
-        return self.name
+        return self.get_name
+
+    @property
+    def get_name(self):
```
punchagan: This is smart. :+1: 

But, using `get_name` as a property seems a little weird. It would be better if we can come up with a nicer name for the property - though, I can't think of anything, right now.

akvo/rsr/static/scripts-src/project-main/project-main-partners.js
```diff
@@ -82,9 +89,42 @@ function renderPartnersTab() {
...
+            /*
+            Return org.long_name if org.name looks like a truncated version of org.long_name
+
+            When importing organisations from IATI, the name field may have been truncated to 25 or
+            (more seldom) 40 characters.
+            Therefore we check if the name is exactly 25 or 40 chars long AND is identical to the
+            substring of long_name to that length. If this is the case we use the long_name as there
+            is a very high probability that the name field has been truncated.
+
+            NOTE: this is also done in the Organisation.get_name method
```
punchagan: I think it would be better to change the organization serializer and return this "clean" name as one field, or even over-write the name field to be this "clean" name.  I'm not sure if overwriting the name field in the json response would cause some confusion, and it is better to keep the serialized form have a one-to-one mapping with the model and it's properties, though

akvo/rsr/models/organisation.py
zzgvh: Ya, ideally it should be "name"...already taken 😛  That's why I settled for get_name.

akvo/rsr/static/scripts-src/project-main/project-main-partners.js
zzgvh: Yup, I was thinking about that too, but it's yet another addition to an already overloaded "quick fix".

zzgvh:
Closing this as it's replaced by #2796.

PR #2783, opened 2017-07-24
[#2782] Too many queries to figure if indicator should aggregate

### LINKED ISSUE #2782: 504 timeouts on the results framework page for project with lot of children and indicators

punchagan:
When an indicator has lots of child_indicators, computing the property
`children_aggregate_percentage` does too many queries. A result with a bunch
of such indicators times out during the call
`/rest/v1/indicator/?result__project=5659`

Closes #2782

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

```
Improve response time for results frameworks with many child indicators [#2782](https://github.com/akvo/akvo-rsr/issues/2782)
```

### DIFF
--- a/akvo/rest/views/indicator.py
+++ b/akvo/rest/views/indicator.py
@@ -12,7 +12,12 @@ from akvo.rsr.models import Indicator
 class IndicatorViewSet(PublicProjectViewSet):
     """
     """
-    queryset = Indicator.objects.all()
+    queryset = Indicator.objects.all().select_related(
+        'result',
+        'result__project',
+    ).prefetch_related(
+        'child_indicators'
+    )
     serializer_class = IndicatorSerializer
     project_relation = 'result__project__'
 
@@ -20,6 +25,11 @@ class IndicatorViewSet(PublicProjectViewSet):
 class IndicatorFrameworkViewSet(PublicProjectViewSet):
     """
     """
-    queryset = Indicator.objects.all()
+    queryset = Indicator.objects.all().select_related(
+        'result',
+        'result__project',
+    ).prefetch_related(
+        'child_indicators'
+    )
     serializer_class = IndicatorFrameworkSerializer
     project_relation = 'result__project__'
--- a/akvo/rsr/models/indicator.py
+++ b/akvo/rsr/models/indicator.py
@@ -196,7 +196,7 @@ class Indicator(models.Model):
         """
         if self.measure == '2' and self.is_parent_indicator() and \
                 self.result.project.aggregate_children and \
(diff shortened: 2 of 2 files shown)


punchagan:
@zzgvh I refactored this PR to change the query slightly, and to avoid prefetching too much data. I like the current code, more then the previous change.

akvo/rest/views/indicator.py
```diff
@@ -12,7 +12,10 @@
 class IndicatorViewSet(PublicProjectViewSet):
     """
     """
-    queryset = Indicator.objects.all()
+    queryset = Indicator.objects.all().select_related(
```
zzgvh: Do the same for the IndicatorFrameworkViewSet maybe? I don't know if it's used anywhere, but it should suffer from the same problem.

akvo/rsr/models/indicator.py
```diff
@@ -196,7 +196,7 @@ def children_aggregate_percentage(self):
         """
         if self.measure == '2' and self.is_parent_indicator() and \
                 self.result.project.aggregate_children and \
-                any([ind.result.project.aggregate_to_parent for ind in self.child_indicators.all()]):
+                any(self.child_indicators.values_list('result__project__aggregate_to_parent', flat=True)):
```
zzgvh: Brilliant! Just one question, would it be even more performant if we prefetched child_indicators since that's the reverse of the Indicator.parent_indicator FK?

akvo/rest/views/indicator.py
punchagan: Yes, I realized I should've done this just after I asked you to look at it. Fixed it now.

akvo/rsr/models/indicator.py
punchagan: The query that runs looks something like this:

```
SELECT "rsr_project"."aggregate_to_parent" FROM "rsr_indicator" INNER JOIN "rsr_result" ON ( "rsr_indicator"."result_id" = "rsr_result"."id" ) INNER JOIN "rsr_project" ON ( "rsr_result"."project_id" = "rsr_project"."id" ) WHERE "rsr_indicator"."parent_indicator_id" = 23156 ORDER BY "rsr_indicator"."order" ASC, "rsr_indicator"."id" ASC
```

Prefetching the `child_indicators` doesn't seem to really change the query in any way.

punchagan: But prefetching the `child_indicators` gets rid of a query when the check `self.is_parent_indicator()` is being made, hence making the overall response slightly faster.  I've added a prefetch for `child_indicators`.

zzgvh: Ya, that was the improvement I thought would happen. Cool!

Issue #2715, opened 2017-06-05
Indicator update design changed to follow approved design


AkvoAnt:
The design/layout of the indicator update to be changed to follow the overall design of that proposed for disaggregation: 

https://projects.invisionapp.com/share/V8B3AI1FH#/screens/226453925

This design needs to be implemented first before proceeding with disaggregation and other design/interaction improvements to the new results framework.

punchagan:
Closed in #2779

PR #2788, opened 2017-07-25
[#2715] Fix hiding when update form is open

zzgvh:
In some cases, when opening the update form, the rest of the tree wasn't hidden correctly.

Fix by calling uiHideMode() when opening the form.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
This PR does fix the problem, but introduces another. When the update form is opened when no filter is in force, on closing the state the accordion was before opening the form is not restored. This leads to the view being being "broken" and you need to either apply a filter or click on the Full view/Overview button to get back to a correct state.

I haven't fixed this because I'm not sure how to fix this. To get back to the state the accordion was in before opening the update form is difficult in the general case. It would mean that we need to record the state of the accordion as it is before opening the form and then restore that state when the form is closed.

Another option is do disallow editing/creating updates when not in one of the filtered states. This is much simpler, but the user looses the ability to "browse" the tree of results and create/edit updates ad hoc.

Issue #2804, opened 2017-08-10
Update buttons disabled when they shouldn't be


zzgvh:
The indicator update buttons do not get enabled when a value is entered as they should. This is because there is no default measure.

punchagan:
Closed in #2805

PR #2805, opened 2017-08-10
[#2804] Fix update buttons disabled

zzgvh:
The indicator update buttons do not get enabled as they should if there is no indicator.measure value.
Since no measure value implies a Unit measure, add it if there is no measure.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


punchagan:
1 line changed, and image :stuck_out_tongue:

Issue #2807, opened 2017-08-10
Simplify qualitative update form


loicsans:

punchagan:
Closed in #2808

PR #2808, opened 2017-08-10
[#2807] Simplified qualitative update

loicsans:


loicsans:
A qualitative update only has the actual value text area:
screen shot 2017-08-10 at 13 46 01

PR #2814, opened 2017-08-11
[#2807] Added a separator when several periods are present

loicsans:
Go to an indicator that has several periods. Check that the periods are separated by a stroke.


loicsans:
Fixing this issue:

PR #2839, opened 2017-08-28
Initial support for disaggregations

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

disaggregation-proposed

The current design adds two new models `IndicatorDimension` and `Disaggregation`.  

An `IndicatorDimension` describes each dimension over which we want the data disaggregated, for example, `Gender-Male` is one dimension, `Gender-Female` is another. 

A `Disaggregation` holds the disaggregated value along a specified dimension, for a specified update, as of now.  Currently, disaggregations are only defined for "actual values", but in future these could also be used for target and baseline values.  Also, `Disaggregation`s could be used to hold disaggregated values over a period, or at the indicator level, also, in future.


zzgvh:
I haven't looked at the code in detail yet, too much happening today...but overall it looks good! Works fine, and I haven't found any large issues.

Some small ones tho 😉 :

  * Entering a string in a disaggregated field results in a 400 API response when saving, but no indication of a problem. I suggest checking it's a number in the client, I think we do that for regular values already.
  * The order of disaggregations is not the same in the update as in the project editor. Eventually we want to be able to reorder them in the editor too I think, but for now having them appear in the same order would be nice.
  * Also the layout could use some love, but I guess that's best left to Loïc.

Issue #2849, opened 2017-09-01
Migrate actual values and comments in periods entered in the project editor to indicator updates


punchagan:
#2666 proposes disabling Actual Value and Actual comment field from the Project Editor.  Existing values need to be migrated to "updates".

PR #2863, opened 2017-09-12
[#2849] Create updates for periods which have been directly updated

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

```markdown
Migrate actual values and comments in periods entered in the project editor to indicator updates [#2849](https://github.com/akvo/akvo-rsr/issues/2849)
```


zzgvh:
Migration looks good! I've added timestamps derived from the LogEntry timestamp. Check it out and merge if it's OK!

punchagan:
@zzgvh The change you made looks okay to me, even though we are using a hack.  Should we also add an internal note on these updates that indicates that these updates have been created during a migration?  Since we are attributing data to a user, based on some heuristic, and creating "back dated" updates, I feel a little uncomfortable making these changes without any note. 

Also, I was wondering if we should have an "RSR System" User for these kind of things.  Some of the indicator periods fall through, in this migration, because we aren't able to find a user.  We could create updates as this user for such periods.  The internal notes mentioned above can also created by this user.  

Thoughts?

punchagan:
@zzgvh This PR is now ready to be reviewed.

PR #2892, opened 2017-09-22
[#2851] Remove the STATUS_NEW_CODE for updates

### LINKED ISSUE #2851: New state on updates should be migrated to draft
There are updates in the DB with the status new. They should be draft.

punchagan:
And migrate all the updates with new status to a draft status.

Closes #2851

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

```markdown
New state on updates should be migrated to draft [#2851](https://github.com/akvo/akvo-rsr/issues/2851)
```


punchagan:
Thanks for the catch, @zzgvh. I had forgotten to add the migration file to the commit. I've fixed the PR, now.

zzgvh:
I did some more digging around the new updates. All except two have the value of 0, and the two that have values are for project 2, our demo/test project. So I suggest we delete all those updates, they are of no use. Plz merge @punchagan if you agree 😄

punchagan:
I think some of these updates have comments, even if the values are '0'.  I didn't check if they had attached files/photos.  I would prefer not deleting the data, unless we are entirely sure all these updates were "useless" updates.

zzgvh:
Ah, didn't think of that, I'll revert my commit.

PR #2893, opened 2017-09-22
[#2822] Minor UI tweak to keep comment label on a single line

### LINKED ISSUE #2822: UI tweaks for the Results framework
There are two small UI issues in the update form:

screenshot_200

  * "says:" falls down to a new line even when there's space enough to keep it on the same line as the name.
  * If we remove "internal" from the internal notes input field the full text will be visible.

punchagan:
Closes #2822

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
I'm pushing a commit that breaks the note in two: the header and the text go on separate lines. I think it looks much better:

screenshot_207

However this reveals another problem with longer notes:

screenshot_208

As you can see the text overflows to the right. I tried to fix this [here](https://github.com/akvo/akvo-rsr/blob/86b67aea2e059f15891a1b9c76afd358e11fca24/akvo/rsr/static/styles-src/main.scss#L5202)  changing it to ```width: calc(100% - 40px);``` but that affects the space available in the form too making it narrower than it needs to be. So I think letting Loïc have a look at it might be the best.

Issue #2946, opened 2017-11-07
More orphaned periods found in the DB


punchagan:
Not sure how, but the DB seems to have orphaned periods - periods whose indicators have parents, but they themselves don't have a parent period.  This causes result aggregation to not work as expected.

Issue #2928, opened 2017-10-17
IATI codelists are not translated


zzgvh:
We use the IATI codelists in a number of places. The lists are generated using [this script](https://github.com/akvo/akvo-rsr/blob/8e8864e00a5cc58870a1303e8dd8f3f3d48c8631/akvo/codelists/scripts/iati_codelist_generator.py), that imports the XML sources from IATI and transform them into python data.

The most common use of the lists are as choices for model fields. We could expand the importing script so that calls to the translation machinery are added to the relevant columns for the lists that are currently in use.

zzgvh:
### Notes on the codelist translations

Currently only a few codelists are set to be tagged for translation, controlled [here](https://github.com/akvo/akvo-rsr/blob/001fc618ee66223f50581294ae51b505db52d717/akvo/codelists/scripts/iati_codelist_generator.py#L32). The full list, including all the currently commented out lists, represents all codelists that are used in RSR.

Some googling reveal that most of those lists already have translations [here](https://github.com/IATI/IATI-Codelists/tree/version-2.02/xml). (Also note that there's a ["sister" repo](https://github.com/IATI/IATI-Codelists-NonEmbedded) that is not as tightly controlled for correctness)

Since the number of strings to translate for the codelists above is fairly small it seems the easiest way to use the translations is to upload the English .po file to Transifex and then just copy-paste the French strings from the XML files in the IATI-Codelists repo into Transifex.

However moving forward, if we want to translate the full list of the codelists used in RSR, it could make sense to write some code to use the translations. Some of the lists are pretty long after all.

Reading the docs on Transifex it seems that if you upload a .po file with translations included, Transifex will update with those translation, even overwriting existing translations. If this is correct (double check, I read quickly) then we could extend the iati_codelist_generator.py script to grab any French translations and export them in the format used in the .po files at the same time as the codelists are assembled. This could be cut'n'pasted into the existing French .po (or have the script inject them if you're fancy!). The updated French .po is then uploaded to Transifex to update it.

#### Note on Transifex
I had some problems when doing the French translation for the Results framework last week. The reason was I initially uploaded the wrong file. I used the French .po file after I had updated it with the newly tagged strings from the results pages etc. However this is not the correct procedure and there were no new translation string to work on.

Instead the file to upload is the **English** "source" .po which only has the original strings and no translations. Transifex then figures out what's missing for the enabled languages and the translation can happily proceed.

PR #2960, opened 2017-11-10
[#2951] Disallow multiple result/indicator/period children per project

### LINKED ISSUE #2951: Updates to parent results framework not propagated to children correctly
Creating a new period/indicator on a parent results framework doesn't create the children periods/indicators correctly, though the code intends to do so.  This either needs to be fixed, or the code removed to avoid breaking things.

punchagan:
Closes #2951, #2928 and #2946

When the results framework is imported from a project to it's children, each
child project should have only one child per result/indicator/period, and no
more.  This PR enforces this constraint in the DB.

This PR also adds a migration to delete offending data.  Luckily, there's not much real data that is conflicting and affected by this bug.  

All the conflicting periods seem to be on a "dummy"/training child project. **NOTE**, we need to confirm that it is safe to delete these results, before doing so. 

There are only a couple of indicators conflicting, and neither of them seem to have any data. 

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry
 ```markdown
Updates to parent results framework not propagated to children correctly [#2951](https://github.com/akvo/akvo-rsr/issues/2951)

Also includes fix to IATI codelists are not translated[#2928](https://github.com/akvo/akvo-rsr/issues/2928) and More orphaned periods found in the DB[#2946](https://github.com/akvo/akvo-rsr/issues/2946)
```

### CODE REVIEW

File: akvo/rsr/migrations/0113_auto_20171110_1042.py [hunk 1]
```diff
@@ -0,0 +1,171 @@
...
...'application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', 'application'), ('application', '
```
zzgvh: Here, and in several other places the choices list is based on the buggy codelist_choices function that's fixed in #2966. 

I don't think this matters, it seems the choices list here is more of an annotation that the real thing since we're generating the choices dynamically via the function call and I can't see the list in the migration ever being used. Just wanna get a second opinion from you.

punchagan: You are right about the values in the migration not being used, but when this branch gets merged with develop (and the fixed `codelist_choices`), django will ask us to migrate, anyway - a new migration will become possible.  Instead, I just rebased this PR against the current develop, and fixed the migration file.

Issue #2973, opened 2017-11-17
Add 'lock' and 'unlock' icon to Indicator Reporting Periods 


AkvoAnt:
Currently it is difficult for a User to know whether or not an indicator reporting period is 'unlocked' or 'locked'. 

It would be to add a 'unlock/lock' icon so that the User could clearly identify the status of an indicator reporting period. 

![lock_unlock icon](https://user-images.githubusercontent.com/26227538/32950732-6b00efe8-cb75-11e7-9146-1741fa125538.png)

PR #2978, opened 2017-11-20
[#2973] Add icon to show lock status for periods

zzgvh:
Return the legend showing if an indicator period is locked or not, but replace the text with an icon.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

```markdown
Add 'lock' and 'unlock' icon to Indicator Reporting Periods [#2973](https://github.com/akvo/akvo-rsr/issues/2973)
```


zzgvh:
Ah, true! Good catch.

punchagan:
> Ah, true! Good catch.

@zzgvh I have pushed a commit to fix this. Please take a look and merge if it looks alright to you.

PR #2992, opened 2017-11-27
Issue/2942 target value comment

loicsans:


zzgvh:
Tooltip looks good, but there's a snag: the wrapper &lt;div&gt; is so wide that it blocks access to the **X** dismiss button rightmost in the form. I tried narrowing wrapper but then the whole tooltip text box gets narrower too 😢

loicsans:
I don't think it matters, since you are not supposed to use the "x" button when you hover the "i" to display the tooltip

zzgvh:
That would be the case if it wasn't for the div being so wide as to overlay the X making it almost impossible to click on it. I tried to make a screenshot of it but the cursor won't be part of the shot making it impossible to show...

Issue #2940, opened 2017-10-25
Bulk approval of indicator updates


AkvoAnt:
Provide functionality that allows M&E managers, for the EUTF project, to ‘bulk’ approve indicator reporting updates due to the fact that the EUTF does not use reporting periods as a typical RSR results framework user. 

Action to approval flow: 

- Add check boxes for each indicator period appearing in the ‘Pending Approval’ tab 
- Enable bulk approving of periods in the ‘Pending Approval’ tab 
- Once indicator reporting updates have been approved (i.e. via bulk approving) they are no longer (visible) in the ‘Pending Approval’ tab.

PR #2993, opened 2017-11-28
Results framework UI changes for EUTF

### LINKED ISSUE #2982: Modify results framework to better suit the EUTF workflow
The EUTF does not use reporting periods as a typical RSR results framework user.  The UI for EUTF can be modified as below, to suit their needs

- Remove/hide the ‘Needs Reporting’ tab 
- Remove/hide the ‘Approved’ tab
- Remove/hide the ‘Filter Period’ menu 
- Remove/hide the ‘Lock’ and ‘Unlock’ buttons

punchagan:
This branch is based off of #2940, and needs to be merged after that is merged. 

Closes #2982

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
Both this and the bulk approval PR look fine, but are all EUTF projects' periods unlocked? Otherwise they just lost the tools to manage that...

punchagan:
@zzgvh Ah, very good catch!  Also, given that by default a newly created indicator period is locked, there needs to be some way to unlock these periods. Any ideas, @zzgvh @AkvoAnt ?

akvo/rsr/management/commands/lock_unlock_periods.py
```diff
@@ -21,28 +21,40 @@ class Command(BaseCommand):
         make_option('-k', '--keyword',
```
zzgvh: I can't get keywords to work. They can be more than one word and even if i wrap them in " only the first token is used: ```./manage.sh lock_unlock_periods unlock -k "EU Trust Fund - North of Africa"``` results in the following debug output:
```
(Pdb) options
{'keyword': 'EU', 'settings': None, 'pythonpath': None, 'verbosity': u'1', 'traceback': None, 'no_color': False, 'organisation': None}
```

punchagan: There are multiple problems here. 

1. The `manage.sh` script does some bash magic, and doesn't handle the args with spaces very well. Escape the spaces, or use `manage.py` from the vagrant box. 
2. They keyword you were trying to use has a unicode hyphen, it looks like `\u2013`.  Try using `'EU Trust Fund - Sahel and Lake Chad'`

akvo/rsr/static/scripts-src/my-results/components/FilterBar.jsx
```diff
@@ -5,31 +5,32 @@
    < http://www.gnu.org/licenses/agpl.html >.
```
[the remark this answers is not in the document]
punchagan: I agree.  As a work around for this time, I ran prettier on all the files changed in this PR, in a separate commit first.  All other commits should be minimal changes.

punchagan:
@zzgvh the points you raise about workflow annoyances are quite valid. Let's have a discussion on this and decide what route to take.

punchagan:
Too old, and probably no longer needed.

PR #2996, opened 2017-11-28
[#2995] Recalculate the country when lat/lng change

### LINKED ISSUE #2995: Updating location's lat/long should update the country field
The country field of a location is only calculated and updated on first save. Changing the latitude/longitude after wards, doesn't update the country. 

Because of how the project editor save works, each field is saved separately, this causes the country computation for locations to be incorrect.

punchagan:
To keep the save fast, previously country was recomputed only when there was no
country set previously. This leaves a user with no way to update the country, if
an incorrect lat/lng had been entered.

Also, given the way project editor's save works correctly, latitude alone is
saved first, and the longitude is updated on the location, in a succeeding step.
This causes an incorrect location to be set on the project first, and that never
gets updated.

Closes #2995

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry
```markdown
Updating location's lat/long should update the country field[#2995](https://github.com/akvo/akvo-rsr/issues/2995)
```


punchagan:
This fix should probably be accompanied by a migration that fixes all the locations with an incorrect country.

punchagan:
@zzgvh Thanks for looking into this PR.  A couple of tests were failing, and I fixed them. 
1. Country was being unset, when no lat/long were set because of default values being (0, 0). 
2. A test used a lat long that didn't really belong to the country it was trying to use.

PR #3033, opened 2017-12-20
Add a markdown editor and markdown rendering for qualitative updates

punchagan:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

This PR is based on #3018, and should be merged after that is merged.


zzgvh:
@punchagan I found a couple of issues.

Both the edit form and the preview are visible in the update form. I suspect this is because of wonky CSS.

The actual text is not rendered in the update, but this is because of the missing {} to destructure the data in the QualitativeUpdateBody parameters.

However when I fixed this some styling is not applied correctly, the bullets for bullet lists are missing. I'm guessing CSS is to blame here too 😉

punchagan:
> Both the edit form and the preview are visible in the update form. I suspect this is because of wonky CSS.

That's weird. I'll take a look at this. 

> The actual text is not rendered in the update, but this is because of the missing {} to destructure the data in the QualitativeUpdateBody parameters.

The develop branch had to be merged into this, but I didn't do this since @loicsans and I were working on the same branch.  I'll do this now. 

> However when I fixed this some styling is not applied correctly, the bullets for bullet lists are missing. I'm guessing CSS is to blame here too 😉

Yes.

akvo/rsr/static/scripts-src/my-results/components/updates/UpdateForm.jsx
```diff
@@ -934,12 +959,27 @@ export default class UpdateForm extends React.Component {
...
-        } else {
-            this.saveUpdate(action);
-        }
+        switch (action) {
+            case c.UPDATE_ACTION_DELETE: {
+                this.deleteUpdate();
+                break;
+            }
+
+            case c.UPDATE_ACTION_EDIT: {
```
zzgvh: I think this case and the next never happen.

akvo/rsr/static/scripts-src/my-results/components/common.jsx
```diff
@@ -169,3 +174,65 @@ export class FileReaderInput extends React.Component {
...
+
+    onChange (reactMde) {
+        const {text, selection} = reactMde;
+        this.setState({reactMde})
+        this.props.onChange(text, selection);
+    }
+
+    render() {
+        const {show_editor} = this.state;
+        const editClass = show_editor?'active': ''
```
zzgvh: Here's a weird one. If I append ";" to this line and the next, as my editor tells me to, the update form preview is always visible. I'm too tired to figure out why now, but maybe you can tomorrow? (If you can reproduce the behaviour...)

screenshot_231

akvo/rsr/static/scripts-src/my-results/components/updates/UpdateForm.jsx
punchagan: Yes, I had reverted a commit, and seems like git revert didn't revert this bit.

akvo/rsr/static/scripts-src/my-results/components/common.jsx
punchagan: I wasn't able to reproduce this. The `editClass` is only used to change how the 'Write' and 'Preview' buttons look, not to control whether preview button is shown or not. 

Can you try running `npm install` before running `npm run watch`?

PR #3056, opened 2018-01-03
Couple of latency fixes when changing indicator update state

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rest/serializers/indicator_period_data.py
+++ b/akvo/rest/serializers/indicator_period_data.py
@@ -73,7 +73,14 @@ class IndicatorPeriodDataFrameworkSerializer(BaseRSRSerializer):
             )
             serializer.update(disaggregation_instance, serializer.validated_data)
 
-        return instance
+        return instance._meta.model.objects.select_related(
+            'period',
+            'user',
+            'approved_by',
+        ).prefetch_related(
+            'comments',
+            'disaggregations',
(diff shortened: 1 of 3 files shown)


akvo/rest/views/indicator_period_data.py
```diff
@@ -37,11 +37,14 @@ class IndicatorPeriodDataFrameworkViewSet(PublicProjectViewSet):
...
         'period',
+        'period__indicator',
+        'period__indicator__result',
+        'period__indicator__result__project',
         'user',
         'approved_by',
     ).prefetch_related(
         'comments',
-        'comments__user'
+        'comments__user',
```
zzgvh: I think we should include disaggregations here too.

akvo/rest/serializers/indicator_period_data.py
```diff
@@ -73,7 +73,14 @@ def update(self, instance, validated_data):
             )
             serializer.update(disaggregation_instance, serializer.validated_data)
 
-        return instance
+        return instance._meta.model.objects.select_related(
```
zzgvh: Good catch!

punchagan: by the tests. But, nevertheless, good. 👍

Issue #3058, opened 2018-01-03
OrganisationIndicatorLabelViewSet inherits from the wrong parent


zzgvh:
**OrganisationIndicatorLabelViewSet** mistakenly inherits from **PublicProjectViewSet** instead of **BaseRSRViewSet** leading to errors [here](https://github.com/akvo/akvo-rsr/blob/1c0592af97e000a830907ed652a3eef1dbd5f6ca/akvo/rest/viewsets.py#L150) since there is no relation to a project.

punchagan:
See also #3053

punchagan:
Closed in #3059

PR #3059, opened 2018-01-03
[#3028] Fix base class of OrganisationIndicatorLabelViewSet

zzgvh:
OrganisationIndicatorLabelViewSet should inherit from BaseRSRViewSet since it has no association with a project.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
This bug shows up if you access the endpoint while not logged in.

zzgvh:
Issue link in the commit heading is wrong, should be to #3058

Issue #2809, opened 2017-08-10
Enforce numerical values for quantitative indicators


zzgvh:
We should not allow non-numerical values for indicators that aren't qualitative. I'm copying what I wrote in #2756 here:

With the introduction of qualitative indicators I would recommend that we enforce numerical values only for quantitative updates.

To see if this would be hard, I checked how many updates we currently have that would need "fixing", i.e. they are updates with non-numerical values. The results look like this

Value | Count
-----|-----
Empty string  | 7
Hyphen (-) | 5
?? | 4
Available | 13
DWSSP guidelines and DWQS formulated | 1
Gender aware | 5
Gender neutral | 3
High | 1
Implementation coordination committee   established | 1
Maintain inclusion | 1
Met | 1
N/A | 110
No monitoring system in place | 1
Not available | 10
Now in progress | 1
Operational | 2
Prototype available | 4
Rural water construction standard   established | 1
See description | 4
TBD | 2
WASH coordinator position established | 1
Yes | 2
all | 1
available | 1
in progress | 4
n.a. | 1
n/a | 3
nd | 1
no | 1
test | 1
yes | 5
using this SQL:
```sql
SELECT
  u.value,
  count(u.value)
FROM rsr_project p, rsr_result res, rsr_indicator ind, rsr_indicatorperiod per,
  rsr_indicatorperioddata u
WHERE u.value !~ '[0-9]' AND res.project_id = p.id AND ind.result_id = res.id AND
      per.indicator_id = ind.id AND u.period_id = per.id
GROUP BY u.value
ORDER BY u.value
```
(The SQL can be easily modded to show which projects, results etc the updates belong to)

The Value column is the text entered as data for the update, the Count column is how many updates have this exact value. Note that the first value is an empty string, and the second is only a hyphen, "-", but markdown mess them up 😉. The results above suggest that it should be easy to migrate the values since most of them are an indication that there is no value. The few that hold any data can be moved to the text field for the update.

To see which projects and organisations that have  non-numerical values in current indicator updates I ran this SQL:

```sql
SELECT DISTINCT
  org.name as Organisation,
  project.id as "Project ID",
  update.value as Value
FROM 
  rsr_project project, 
  rsr_result result, 
  rsr_indicator indicator, 
  rsr_indicatorperiod period,
  rsr_indicatorperioddata update, 
  rsr_partnership partnership, 
  rsr_organisation org
WHERE
  update.value !~ '[0-9]' AND 
[...]

PR #3076, opened 2018-01-08
Issue/2809 migrate non numeric indicator values

zzgvh:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/management/commands/check_non_numeric_indicators.py
+++ b/akvo/rsr/management/commands/check_non_numeric_indicators.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+
+# Akvo Reporting is covered by the GNU Affero General Public License.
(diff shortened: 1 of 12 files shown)


zzgvh:
The PR contains both management scripts that search for non-numericals baseline, target and actual values and migrations that migrate those non-numeric values to comment fields.

The reason for the "duplication" is that I started out creating the management scripts so we could see how many values and what kind of values there were to migrate.

As it became clear that the values, particularly the target values were so many that we had to solve this with code, I used the management scripts as a base for the migrations.

Note that this is the second branch that refers to #2809, the first one was based on a very old develop and had clashing migrations. It was easiest to make a new branch.

zzgvh:
@puneeth seems I've managed to break a bunch of tests. I forgot to run all tests, I only ran the new tests I made 😛 If you have the time to check them please do, otherwise I'll have a look in the morning.

PR #3103, opened 2018-01-23
Handful of header and footer related changes

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
### General thoughts on the new layout

I know that this is a work in progress, so I might comment on things that won't remain that way.

Currently our listing pages (projects, organisations, updates) have no headers telling the user what they see. Now that those headings are gone from the top bar, you're completely lost as to what you see on the listing pages. I know this is a responsive layout so it's not true for small mobile devices, but apart from that there's room for a title to the right of the RSR logo telling you where you are.

When logged in the top bar is populated with more stuff. I can't see why not all of those links can't be in the same place, probably in the "logged in select" with your name. This is particularly true when the hamburger manu's shown for smaller screens. Now we have the hamburger and inside it the logged in select.

BTW using the name in the logged in select is not ideal since there's no guarantee that the user has a name! (Yes, this is incredibly silly...)

Why did the My IATI link get singled out for moving from the rest of MyRSR?

zzgvh:
### UI bugs

UI issues big and small!

Buttons have different size (Firefox):

screenshot_234

Strange scroll position and advanced search visible on loading (Chrome):

screenshot_235

Advanced search is on top of hamburger menu:

screenshot_236

Language select should open "upwards". Currently cou have to scroll down after clocking on it to see all of it:

screenshot_237

screenshot_238

Search field in user management too wide at certain widths of the window:

screenshot_239
screenshot_240

Toggle button switching between live and admin view is centered in the hamburger menu (Chrome):

screenshot_241

Language chooser and background aren't happy together when window is narrow (Chrome):

screenshot_242

Hamburger text disappears when hovered over (Chrome):

chrome

zzgvh:
### New bugs

Found in all browsers (Chrome, FF and Safari).

Vertical align not correct in the top bar:

screenshot_245

The user admin select becomes only an arrow when the user has no first and last name. I mentioned this already, here's an example:

screenshot_246

akvo.org logo not centered in "hamburger view":

screenshot_247

Issue #3111, opened 2018-01-29
Misaligned target value comment tooltip


loicsans:

punchagan:
Closed in #3112

PR #3112, opened 2018-01-29
[#3111] Target value comment realigned

loicsans:


loicsans:
screen shot 2018-02-08 at 14 47 08
I cannot reproduce what you show here, for me it's aligned

punchagan:
Not able to reproduce this, either. May be @zzgvh had old code running? Merging this for now

We can push new code, if there's another bug which Gabriel ran into.

PR #3154, opened 2018-03-02
Make project directory loading and filtering more dynamic

punchagan:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


punchagan:
- [ ] Think about how this effects SEO
- [ ] There are some tests for API responses on partner sites. Check if the UI works fine too.

zzgvh:
I found a bug. 

Go to an organisation page via the link for main partner in a project. On the org page select one of the links back to the project listing page under the Projects in RSR heading, e.g. IMPLEMENTATION.

Now in some cases, it's a little hard to provoke, when you click the reset filter link, all selects get populated with undefined objects. I can provoke it most of the time if I open one of the other selects and then click reset.

screenshot_258

zzgvh:
Layout problem:

screenshot_259

I've seen this on test too where there are actual photos so it's not because I don't have pics locally.

zzgvh:
I think there's a need for some feedback when the search is executed. On a slow machine, like my local, it takes quite a while before the filter kicks in and there is no visual indication that something about to happen.

punchagan:
> I think there's a need for some feedback when the filtering is executed. On a slow machine, like my local, it takes quite a while before the filter kicks in and the only visual indication that something might be happening is that the selects get grayed out.

I've added a spinner for this

image

punchagan:
> I've seen this on test too where there are actual photos so it's not because I don't have pics locally.

Yes, I have seen this too.  I'll take a stab at this, but it's best if @loicsans looks at it.

punchagan:
Closing in favor of #3197

Issue #3177, opened 2018-03-15
Refactoring and simplification of toggle view functions in my results


emej:
Refactoring and simplification of toggle view functions for the Full view/Overview button in my results

PR #3179, opened 2018-03-19
[#3177] added new function for toggle view and removed old ones

emej:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


punchagan:
https://github.com/akvo/akvo-rsr/pull/3179/files?w=1 to hide whitespace changes while reviewing this.

akvo/rsr/static/scripts-src/my-results/utils.js
```diff
@@ -286,91 +286,67 @@ export function fullUpdateVisibility(update, activeFilter) {
...
-    };
-    const children = node.children.filter(child => child.model !== undefined);
-    const childKeys = children.map(node => keysList(node, open));
-    return flatten([key].concat(childKeys));
-}
-
-export function toggleTree(model, id, open, MEManagerKeys) {
-    const fullTree = tree(model, id);
-    return keysList(fullTree, open, MEManagerKeys);
+export function toggleTree(open, MEManagerKeys) {
```
zzgvh: The MEManagerKEys parameter is not needed any longer. Also check for calls and make sure they don't include this param.

akvo/rsr/static/scripts-src/my-results/utils.js
```diff
@@ -286,91 +286,67 @@ export function fullUpdateVisibility(update, activeFilter) {
...
 export function openResults(activeKey, isMEManager) {
-    if (isMEManager) {
+    if(isMEManager) {
         return isMEManagerDefaultKeys(activeKey);
     } else {
         return activeKey == undefined || activeKey.length == 0;
     }
 }
 
-export function createToggleKeys(parentId, model, activeKey) {
```
zzgvh: If you search for createToggleKeys in the code you'll find that it's present in a number of toggleAll() methods. However all of those methods are dead code, they're not called from anywhere so they can be removed too! Also make sure to remove createToggleKeys from the import statements and toggleAll from the constructors.

PR #3189, opened 2018-03-22
[#3188] Fix default periods

### LINKED ISSUE #3188: Analyze the results framework for data integrity problems
I've created a script that analyzes the results framework data for a number of potential data integrity problems. The seed for this is that we are seeing bugs when creating data in the results framework that seem to come from the structure of existing data not being consistent.

Also there are known issues with parent-child relations between projects, and I wanted to understand the extent of this.

zzgvh:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

Closes #3188


akvo/rsr/management/commands/data_integrity.py
```diff
@@ -0,0 +1,181 @@
...
+        'Project ID',
+        'Indicator ID',
+        'Parent indicator result ID',
+        'Indicator result parent result ID',
+        'Indicator title',
+        'Project title',
+        'Result ID',
+        'Result title',
+    ]
+    for indicator in Indicator.objects.all().select_related(
```
punchagan: This loop can be made much more efficient by moving the checks into the DB, and reducing the number of items we loop over. 

```python
for indicator in Indicator.objects.exclude(parent_indicator=None)\
                                  .exclude(parent_indicator__result__pk=F('result__parent_result__pk'))
```

akvo/rsr/management/commands/data_integrity.py
```diff
@@ -0,0 +1,181 @@
...
+        'Project ID',
+        'Project title',
+        'Result ID',
+        'Result title',
+        'Indicator ID',
+        'Indicator title',
+    ]
+    print "Indicator periods where parent indicators don't match"
+    print problem_periods.export('csv'),
+    # Note: this double loop is to avoid running out of memory (I think). Previously the loop was
```
punchagan: I think the double loop can be gotten rid of, by moving the if condition checks into the DB. 

```python
for period in IndicatorPeriod.objects.exclude(parent_period=None)\
                                     .exclude(parent_period__indicator__pk=F('indicator__parent_indicator__pk'))
```

akvo/rsr/management/commands/data_integrity.py
zzgvh: Ah, yes. Nice!

PR #3201, opened 2018-03-28
[#3200] Support CSV format in the REST API

### LINKED ISSUE #3200: Support CSV format for the API end-points
Currently, Lumen supports only CSVs since they are easy to translate to the 2D datasets that Lumen uses. It is tricky to support any arbitrary JSON format, since the translation to a 2D data structure automatically would be hard. 

But, to be able to use RSR data in Lumen visualizations it would be useful to support CSV format using the [django-rest-framework-csv](https://github.com/mjumbewu/django-rest-framework-csv) package, which provides a renderer similar to the json and xml renderers that we already have enabled.

punchagan:
Closes #3200

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry


punchagan:
Here is a [sample CSV](https://github.com/akvo/akvo-rsr/files/1856032/project-updates-2.xlsx) (saved as xlsx to upload to GitHub)

zzgvh:
Seems newlines (\n) in text fields breaks the output. See for example /rest/v1/project/7005/. Messy!

punchagan:
@zzgvh Does it really break the output? It looks fine to me, the fields with `\n` seem to be quoted, and the right thing seems to happen for me.  If you tried to view the CSV in Excel or Libre Office, may be conversion from CSV didn't handle `\n` correctly?

zzgvh:
You, you're correct @punchagan ! It's Excel that is incredibly stupid when importing CSV with line breaks.

### How to import CSV into Excel

Here's a recipe for how to import a CSV with linebreaks into an Excel sheet on a Macintosh (Excel for Mac 2011):

In this example I'm using data from the Akvo test project https://rsr.akvo.org/rest/v1/project/2/?format=csv

- Open the downloaded CSV in TextEdit
screenshot_288

- Open a new Excel sheet
- Copy all of the text (Cmd-A to select all) from TextEdit into the top left cell of the sheet
screenshot_289

- Switch to the Data tab in the Excel toolbar, click on Text to columns
screenshot_290

- In the wizard, choose Delimited in step one, Comma as delimiter in step two. Click Finish
screenshot_291

screenshot_292

- The data will move around a bit, but will still not be put in the correct cells
- Go back to the Home view
- Select all data (Cmd-A or click the space top left of the sheet itself with an arrow facing down and to the right)
screenshot_294

- Click the Clear button and choose All. You should now have what looks like an empty sheet. But it's not quite empty, beacuse it remembers where to put items from the CSV...
screenshot_296

- Paste again. This should paste the text from TextEdit and it'll magically put the data into the correct cells. Ain't Excel great?!? 👿  
screenshot_297

PR #3206, opened 2018-04-02
[#3142] Change the view permissions check to be lax

### LINKED ISSUE #3142: View for projects with no results framework but using RSR Updates loads with error
For the redesigned MyRSR ... Some projects only use RSR Updates, i.e. they do not use the Results Framework. When you click/open a project to view in the new MyRSR you get this message: 

no results framework

For these projects the view should look like this: 

rsr udate view

punchagan:
The decision of showing or hiding a particular tab, is made based on specific
permissions the user has, rather than preventing the whole page from loading.

Closes #3142

  - Log in as an ordinary user (belongs to `Users` group only), and verify that the `/en/myrsr/my_project/<id>` page opens for a project with the user's org as a partner.  
  - Once the page loads, verify that only the reports tab and the add update tab are shown. 

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rsr/static/scripts-src/my-results/components/App.jsx
+++ b/akvo/rsr/static/scripts-src/my-results/components/App.jsx
@@ -348,6 +348,7 @@ export default class App extends React.Component {
             );
         }
         const show_reports = page.mode && page.mode.show_narrative_reports;
+        const show_results = page.mode && page.mode.show_results;
         const projectId = dataFromElement("project").id;
         const has_results = dataFromElement("project").has_results;
         const results_tab = (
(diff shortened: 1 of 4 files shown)


punchagan:
Note to self: The edit project button should be hidden if the user doesn't have permissions. Showing a 403 after click is bad UX.

zzgvh:
🐛 

I have a test user that is a User for Akvo and a Project editor for IRC:

screenshot_268

But when I open my_project for an Akvo project I see the Results tab:

screenshot_269

punchagan:
The fix for additional permissions to users that was hot-fixed was missing on this branch. I pushed a commit to change this.

PR #3208, opened 2018-04-02
Issue/3205 enumerator role

punchagan:
**NOTE**: Based on top of #3206.  That PR can be ignored, if this is merged before that. 

  - Add an enumerator to an organisation
  - Login as the enumerator and verify that they can add/edit/delete project updates. 
  - They should also be able to post updates and generate reports (and do everything else that 'normal' users can do). 

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


punchagan:
@zzgvh this PR is now ready to be reviewed!

punchagan:
This will need to come along with documentation, though. Since, the Enumerator role will be shown to all organisations (in the User management -> invite dialog, for instance), there needs to be documentation on the permissions for this role.  Also, #1989 seems useful to implement.

punchagan:
@zzgvh Nice catch! I fixed this on the UI branch - in [this commit](https://github.com/akvo/akvo-rsr/commit/3a834ac41775b4bf21d9f26086cb258fa574de20).  I'm not entirely sure why this should happen, but it is fixed.

zzgvh:
Maybe we should we merge this branch into the UI branch instead then?

PR #3215, opened 2018-04-11
[#3213] Script removing indicator period duplicates

### LINKED ISSUE #3213: Script to remove duplicate indicator periods
In the project editor it's possible to create indicator periods with exactly the same period_start and period_end as on already existing. This can happen if an indicator is set as default for the project. When you do this you get the question if the default indicator periods should be copied to all other indicators of the project. If you answer yes this is done even if a period with the date range already exists for the indicator. This is a bug that should be fixed, but we also need a management script to remove existing duplicate indicator periods.

zzgvh:
Management script that deletes indicator periods that share the same indicator, have the same
period_start and period_end and has no other data. There will always be one period left
with a particular set of dates.

The script will also list periods that are duplicates as per above, but have data added to
them

Closes #3213 

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
Closing this PR as the code is now in #3188

Issue #3205, opened 2018-03-30
Add an Enumerator role


punchagan:
As @AkvoAnt described it -- It is the current 'User' role +  'Add indicator update', that's it.

PR #3218, opened 2018-04-12
[#3205] Fix subtle bug with user & enumerator permissions for None

punchagan:
The permission predicates for is_org_user and is_org_enumerator were returning
True for None objects without checking if the user has a corresponding
employment of that kind. This isn't a problem for the User group, because users
in the User group have the least permissions, so this bug didn't really manifest
anywhere. When the enumerator role is added, though, users in the Users group,
but not in the Enumerators group will get permissions to add indicator updates,
for example.

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/tests/test_permissions.py
+++ b/akvo/rsr/tests/test_permissions.py
@@ -192,6 +192,9 @@ class PermissionsTestCase(TestCase):
         # None object permissions
         self.assertTrue(user.has_perm('rsr.view_project', None))
 
+        # Test no permissions on None, if user doesn't have required employment
+        self.assertFalse(user.has_perm('rsr.add_indicatorperioddata'))
+
         # Project Update permissions
         for i, project_update in enumerate(self.project_updates):
             test = self.assertTrue if i == 0 else self.assertFalse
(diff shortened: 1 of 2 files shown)


akvo/rsr/tests/test_permissions.py
```diff
@@ -192,6 +192,9 @@ def test_org_user_permissions(self):
         # None object permissions
         self.assertTrue(user.has_perm('rsr.view_project', None))
 
+        # Test no permissions on None, if user doesn't have required employment
+        self.assertFalse(user.has_perm('rsr.add_indicatorperioddata'))
```
zzgvh: This test fails when it's called from test_org_enumerator_permissions() 😢

punchagan: That is the intended behaviour.  As I understand it, checking for a permission on None, is essentially only checking if the user has an employment of the given role, or not.  And, enumerators have the permission to add indicator period updates.

punchagan: Before the fix, the predicates were not checking if the user had the given kind of employment or not, and this was the bug the commit tries to fix.

punchagan: Ah, sorry, I didn't realize the test was calling another test. Bad test habits, I guess. I'll fix the test.

Issue #3278, opened 2018-06-19
Single period hierarchy for EUTF has duplicate periods


zzgvh:
For reasons that are not fully understood, some of the indicators of the EUTF indicators result have two periods. This needs to be fixed ASAP.

zzgvh:
The commit above allows us to run the management script and make sure that any periods with a target value, or any other data, is never deleted. Running it on the current dataset seems to fix all problems in the EUTF hierarchy.

zzgvh:
I've noticed that when running the management script, sometimes new periods are crated even if the dry-run didn't indicate that would happen. Thus the script should be run multiple times until all periods are logged as modified.

PR #3279, opened 2018-06-19
[#3278] Update singe_period_indicators management script

zzgvh:
Change the behaviour slightly when more than one period is encountered.

If there's more than one period, check if only one if "modified". If so keep the modified period and delete the others.

A period is modified if one of the fields target_value, target_comment, actual_value or actual_comment has a value or the period has at least one update.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rsr/management/commands/single_period_indicators.py
+++ b/akvo/rsr/management/commands/single_period_indicators.py
@@ -40,6 +40,23 @@ class Command(BaseCommand):
                 period_start=PERIOD_START, period_end=PERIOD_END
             )
 
+        def find_periods_with_data(periods):
+            """ Separate a list of periods into two groups: those tht have been modified from an
+            original, "empty" state and those that haven't. To be counted as modified a period has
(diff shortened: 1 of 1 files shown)


akvo/rsr/management/commands/single_period_indicators.py
```diff
@@ -40,6 +40,23 @@ def get_parent_periods(indicator, PERIOD_START, PERIOD_END):
...
+        def find_periods_with_data(periods):
+            """ Separate a list of periods into two groups: those tht have been modified from an
+            original, "empty" state and those that haven't. To be counted as modified a period has
+            data in at least one of the fields target_value, target_comment, actual_value or
+            actual_comment or has at least one update.
+            """
+            modified_periods = []
+            unmodified_periods = []
+            for period in periods:
+                updates = IndicatorPeriodData.objects.filter(period=period)
```
punchagan: I would change this to check for `updates.exists()` since `if updates` is much slower than `if updates.exists()`

Issue #3277, opened 2018-06-14
(re) Introduce the 'Edit' function to Project Updates 


AkvoAnt:
After feedback from both partners and the Akvo support team it's been decided to allow Project Updates to be edited. 

Key things consider when creating a solution: 

- Editing permission should be restricted to the creator of the update and to the Admin of the org only;
- Some kind of record of the original update should be kept in the DB in the event an update has been edited/modified; 
- If possible, it would be great if Akvo Super Users were able to both edit _and_ delete Project Updates, (at the request of the partner only) _inside_ RSR (i.e. not in the old admin.)

punchagan:
Closed in #3282

PR #3286, opened 2018-06-26
[#3285] project updates: Send credentials with fetch requests

### LINKED ISSUE #3285: Project updates for private projects don't show up in the project update CRUD page
The problem is with how fetch works in Firefox vs. Other browsers.  Firefox (my default browser) sends credentials in a fetch request automatically, without specifying the credentials: same-origin option to the fetch call - that is the default. In Chrome, the option needs to be specified. 

For private projects, updates are not shown 'publicly'. So, if you don't send credentials, you are treated as an anonymous/public user, and don't see any updates.

punchagan:
Fixes a problem with project updates not being listed for private projects,
since the credentials for the API call is not sent by default in some browsers.

Closes #3285

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
Closing this since the fix is now included in #3277

PR #3317, opened 2018-07-25
LISENCE CORRECTION

beaglebao:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
Thanks for the reminder to update our copyright dates. Apart from that there is nothing in this PR we can use. And we never merge directly into master outside our normal workflow.

PR #3343, opened 2018-07-30
Remove old migrations in RSR

punchagan:
The PR remove migrations for the `rsr` app and the `codelists` app. 

The data migration code that populates codelists has been moved into the slow tests which
actually use this data for cross-checking the API response data, and ensuring the API isn't broken

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


zzgvh:
Looks good! So nice to get back to square one! 

One idea. When you run the migrations you get this question on removing stale content types. It'd be really nice if this could be answered in the affirmative automatically when the deployment machinery runs 😉 This may be outside this PR, if it is go ahead and merge!

punchagan:
I think that is outside this PR, though, I'm not sure what is the right way to do it. `--noinput` doesn't seem to do what we want. I'll investigate this further.

punchagan:
I'm actually leaning towards keeping the migrations and only removing the slow migrations, if possible. I don't know if that makes sense, though.

Issue #3363, opened 2018-08-22
Move new project book keeping to the backend


punchagan:
Creation of project makes multiple API requests, apart from the request to create a project. 

- Adding a reporting organisation
- Adding a log entry for the project creation
- Setting up custom fields. 

This should all be moved to the backend.

PR #3364, opened 2018-08-22
[#3363] Move the project creation book-keeping API calls to backend

punchagan:
Creation of a project makes multiple API requests, apart from the actual request
to create a project.

- Adding a reporting organisation to the project
- Making a LogEntry for the project creation
- Setting up custom fields for the project

These can be moved to the backend to make things faster on the client side.

- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry

### DIFF
--- a/akvo/rsr/models/project.py
+++ b/akvo/rsr/models/project.py
@@ -9,7 +9,7 @@ For additional details on the GNU license please see < http://www.gnu.org/licens
 from decimal import Decimal, InvalidOperation
 
 from django.conf import settings
-from django.contrib.admin.models import LogEntry
+from django.contrib.admin.models import LogEntry, ADDITION
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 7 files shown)


akvo/rsr/models/project.py
```diff
@@ -1593,6 +1593,64 @@ def update_parents(self, update_period, difference, sign):
...
+    def new_project_created(cls, project_id, user):
+        """Hook to do some book-keeping for a newly created project.
+
+        *NOTE*: This hook cannot be moved into a post-save hook since we need
+        information about the user who created this project, to perform some of
+        the actions.
+
+        """
+        # Set reporting organisation
+        organisation_ids = user.approved_employments().values_list('organisation_id', flat=True)
```
zzgvh: I think this should be limited to orgs where we know the user can create projects.

This is of course part of the larger problem of the org being selected essentially randomly, but let's not deal with that for now.

punchagan: Updated the PR to set only orgs where the user can create a project as the reporting org. Nice catch!

Issue #3357, opened 2018-08-14
Password enhancements


zzgvh:
Nuffic requires a number of enhancements around account passwords and management. The following is a high level list of the new requirements.

 1. Passwords are safely stored with the use of a per user salt of 64 bits or more; with the use of a bcrypt hashing algorithm

 2. Minimum of 8 characters long and is max. 2 years valid

 3. The password needs to be a combination of the following 3: capital, number, symbol

 4. User get a 1-time usable password and have to change it to their personal password after logging in

 5. Login attempts will be logged and can be looked into / used when they’ve detected a security thread

 6. After 3 login attempts a user is shut out. The user can request a new password after that. If it is not working an Admin should be able to create a new password / give new access to the user.

zzgvh:
These requirements need to be in production by September 20

zzgvh:
# Discussion on the bullet list above

We need a validator that checks the length and required character components.

The requirement to change the password on first log in is moot since users either create their own account and thus select their own password, or the account is created for them without a password and they have to supply it when activating.

However we might look at unifying this into one workflow. If we don't ask for a password when an account is created, but instead when it's activated, the process will be the same for users that register on their own and users that are invited.

Also this will tie into updating a password after two years, since I think we can just expire the activation to get the user to enter a fresh password. Do we want to store the old hashes to prevent the user from using the same password?

PR #3373, opened 2018-09-17
[#3357] Set minimum password length to 12

punchagan:
- [x] Copyright header
- [x] Code formatting
- [x] Documentation
- [x] Change log entry


punchagan:
Good catch! I've fixed that.

Issue #3414, opened 2018-11-08
Some content owned organisations don't appear in the list of content owned orgs


punchagan:
`EU Delegation Senegal` is missing from the list of `EUTF` content owned organisations

punchagan:
Closed with #3429

PR #3415, opened 2018-11-08
[#3414] Fix circular content ownership bug

punchagan:
When two organisations set each other as content_owner using the direct
`ForeignKey` relationship, an incorrect query actually left out such
organisations from each others' list of content owned organisations.

This commit corrects the query to filter out the correct organisations, but also
fixes a problem with recursive calls which didn't happen previously because of
the buggy query.

The tests also include a tests for another possible bug in the list of content
owned organisations. The fix for this bug can come in a different fix.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rsr/models/organisation.py
+++ b/akvo/rsr/models/organisation.py
@@ -306,10 +306,15 @@ class Organisation(TimestampsMixin, models.Model):
             from .employment import Employment
             return Employment.objects.filter(organisation__in=self).distinct()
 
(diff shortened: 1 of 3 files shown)


punchagan:
I've spent some time breaking my head against the failing test, but haven't been able to figure out why it fails. I am unable to reproduce the failure locally.

punchagan:
Super-seeded by #3429

Issue #3422, opened 2018-11-15
API endpoint to trigger Project.import_results()


zzgvh:
Add an endpoint to the API that calls the Project.import_results() method.

Path: /rest/v1/project/<project ID>/import_results/ Method: POST

zzgvh:
After more discussions with Philibert we need to add an endpoint that imports a particular indicator.

zzgvh:
This change should actually be reverted!

zzgvh:
This is undergoing tests by Philibert right now. Leaving open.

zzgvh:
As a precursor to this, I changed the parent_... fields on the IndicatorSerializer and the IndicatorPeriodSerializer. That PR should be reverted: #3416

PR #3423, opened 2018-11-15
[#3422] Modify API endpoint triggering results import

zzgvh:
Modify the endpoint to return HTTP status codes 201 and 400 for success and failure respectively, and refactor the client side code to determine action depending on the response status.

Also check that the user can change the project results are being imported to and modify the returned data some.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rest/views/project_editor.py
+++ b/akvo/rest/views/project_editor.py
@@ -39,7 +39,7 @@ from django.http import HttpResponseForbidden
 from django.utils.encoding import smart_unicode
 from django.utils.translation import ugettext_lazy as _
 
-from rest_framework import status
+from rest_framework import status as http_status
 from rest_framework.decorators import api_view, permission_classes
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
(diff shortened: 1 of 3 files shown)


akvo/rest/views/project_editor.py
```diff
@@ -810,12 +810,20 @@ def project_editor_import_results(request, project_pk=None):
     project = Project.objects.get(pk=project_pk)
     user = request.user
 
-    if not (user.is_superuser or user.can_import_results()):
+    if not (user.is_superuser or
+            user.can_import_results() and user.has_perm('rsr.change_project', project)):
```
punchagan: `user.can_import_results` seems to be doing something that doesn't make a lot of sense - verifying if the user has at least one `Admin` or `M&E manager` employment.  Just checking for `rsr.change_project` permissions seems like a much better way of doing this, but also gives the permission to Project Editors. It may be worth adding another permission `rsr.can_import_results` instead, and use that.

akvo/rest/views/project_editor.py
```diff
@@ -39,7 +39,7 @@
 from django.utils.encoding import smart_unicode
 from django.utils.translation import ugettext_lazy as _
 
-from rest_framework import status
+from rest_framework import status as http_status
```
punchagan: Looks like `status` is being used in L#944 of this file.

punchagan: I fixed this!

PR #3428, opened 2018-11-30
Create EUTF shadow organisations

zzgvh:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

Note that this branch includes an earlier version of the mgmnt script that set up restrictions for the EUTF hierarchy non content owned orgs. This script was used as the base for eutf_shadow_organisations.py.

Problems with the current data has been discovered that need to be addressed to make this change complete, in terms of EUTF control of their projects. This should be addressed outside of the code however (I think...).

The script emits info on all organisations that get shadows and all users that acquire employments with the shadows. The output can be copy-pasted directly into Excel.

Use the output info to check that EUTF projects now have shadows as partners instead of originals. All new employments should be adminable by an EUTF admin user.


akvo/rsr/models/user_projects.py
```diff
@@ -74,42 +78,26 @@ def check_valid_permission_change(admin, user, projects):
...
         enable_restrictions=True
     )
     if not orgs_with_restrictions.users().filter(id=user.id).exists():
-        raise(InvalidPermissionChange)
+        raise InvalidPermissionChange
 
 
-class InvalidPermissionChange(Exception):
-    pass
+EUTF_ORG_ID = 3394
```
punchagan: Unused variables here that can be removed.

akvo/rsr/models/organisation.py
```diff
@@ -175,6 +175,9 @@ def org_type_from_iati_type(cls, iati_type):
         'self', null=True, blank=True, on_delete=models.SET_NULL,
         help_text=_(u'Organisation that maintains content for this organisation through the API.')
     )
+    # pointer to original org if this is a shadow. Used by EUTF
```
punchagan: I think the comment here could go into the `help_text`

akvo/rsr/models/user_projects.py
```diff
@@ -74,42 +78,26 @@ def check_valid_permission_change(admin, user, projects):
...
+EUTF_PROJECT_ID = 4401
 
 
-class RestrictedUserProjectsByOrg(models.Model):
```
punchagan: Nice catch!

akvo/rsr/models/organisation.py
zzgvh: Heh, I know I thought about that when I wasn't at the computer. Then I forgot to put it there...

Issue #3434, opened 2018-12-06
RSR is very slow


dlebrero:
Loading the project pages takes several seconds

PR #3436, opened 2018-12-06
Faster project pages

dlebrero:
Faster project pages

After analyzing the performance of the project pages (the most expensive ones in RSR),
we noticed that more than 70 DB calls are required to render them.

Some of the queries are very expensive to run or process (more than one second) and
the same queries are done multiple times (up to 9 times in some cases).

Reducing the number of queries by:

1. Using the same updates slice, as creating a new slice means that updates are calculated
several times, which is very slow as it has to pull the
information for the user several times, which is specially slow (up to 1 second)
1. Load partnerships and orgs in one go

This reduced the number of DB calls to around half and makes the project pages around x2 as fast.

Testing done by comparing the output of RSR with and without the patch (see scripts/load-test/compare-all-pages.sh script).

Included a bunch of scripts used to analyse and run the performance tests. Very crude scripts that will need some love if we want to make them more generic

Fixes: #3434


punchagan:
Manually merged the perf improvements on to `develop` and k8s related changes onto the `dockerize-django` branch.

PR #3451, opened 2019-01-09
Issue/3450 dev env crawled by google

dlebrero:
Defaulting to do not allow crawlers, so that we do not need to reconfigure the dev and uat environments.

Enabling with a configuration the crawlers in production. 

Also, just for k8s, adding an environment specific robots.txt file.


dlebrero:
@punchagan Can you please deploy to the dev env?

Issue #3476, opened 2019-02-07
Backend failing liveness probe


dlebrero:
The backend container is being killed about twice a week by k8s because it is failing the liveness probe.

There is no clear pattern of why this is happening. Sometimes is due to some Ruby script doing calling a very expensive url, but not always.

punchagan:
Merged to develop.

PR #3477, opened 2019-02-07
Backend fails liveness probe in production

dlebrero:
Being more lenient by changing the liveness threshold to 2 mins.

Also, added an additional worker so we can handle more concurrent requests

See #3476

### DIFF
--- a/ci/k8s/deployment.yml
+++ b/ci/k8s/deployment.yml
@@ -79,12 +79,16 @@ spec:
             port: 8000
           initialDelaySeconds: 30
           timeoutSeconds: 1
+          failureThreshold: 12
+          periodSeconds: 10
         readinessProbe:
           httpGet:
             path: /healthz
             port: 8000
           initialDelaySeconds: 30
           timeoutSeconds: 1
+          failureThreshold: 12
+          periodSeconds: 10
         volumeMounts:
             - name: "secret-config"
               mountPath: "/var/akvo/rsr/code/akvo/settings/42_django.conf"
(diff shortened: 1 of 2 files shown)


ci/k8s/deployment.yml
```diff
@@ -79,12 +79,16 @@ spec:
             port: 8000
           initialDelaySeconds: 30
           timeoutSeconds: 1
+          failureThreshold: 12
+          periodSeconds: 10
```
orifito: This is the default value

ci/k8s/deployment.yml
```diff
@@ -79,12 +79,16 @@ spec:
             port: 8000
           initialDelaySeconds: 30
           timeoutSeconds: 1
+          failureThreshold: 12
```
orifito: Maybe too high? It'll take 2 mins for an unhealthy pod to be restarted, and there is only one. Default is 3, I would set something like 6.

ci/k8s/deployment.yml
dlebrero: Correct, added it for those of us that do not know the defaults by heart

ci/k8s/deployment.yml
dlebrero: The last liveness issue was due to some requests taking more than 60 seconds, so we want to give at least more than that amount of time. 

Given that up until now the issue has been due to being too stringent, I prefer the 2 mins and see how it goes.

punchagan: Seems reasonable to me.

PR #3478, opened 2019-02-11
Change Project Editor save to save all fields of an object together

punchagan:
The main changes are:

- A bunch of tests for saving in the project editor in various scenarios
- Pulled out the save utility functions into a utility module outside the view module
- Fixed the save to take into account all the fields on an object, instead of saving field by field. 
- Got rid of the 4 time looping over the keys in the data, to ensure that all the submitted data has been processed.  Instead, process the data based on the model hierarchy - Project fields before Results before Indicator before Period, etc.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


akvo/rest/views/project_editor_utils.py
```diff
@@ -0,0 +1,660 @@
...
+                    Country.DoesNotExist, OrganisationIndicatorLabel.DoesNotExist) as e:
+                errors = add_error(errors, e, key)
+                return None, errors
+        else:
+            return None, errors
+
+    # Keywords is the only ManyToManyField
+    if isinstance(model_field, ManyToManyField):
+        try:
+            return Keyword.objects.get(pk=int(data))
```
zzgvh: I think this should be
```
return Keyword.objects.get(pk=int(data)), errors
```
right?

punchagan: yes, that seems right!

PR #3515, opened 2019-03-20
3250 layout issues

martinchristov:
Issue fixed


zuhdil:
Looks good.

punchagan:
I've a 1920 x 1080 monitor and I see the white background from the Project search option be longer than the search bar itself. 

image

On the live server, it looks like this.
image

Issue #3534, opened 2019-04-09
Map not appearing


zuhdil:
The map in the project directory will not appear if there is one project that doesn't have the latitude & longitude attribute.

punchagan:
Closed in #3535

PR #3535, opened 2019-04-09
[#3534] Fix map not appearing

zuhdil:
- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


punchagan:
Good catch! Looks good to me. Do you think we need to fix the `ProjectUpdateDirectorySerializer` too?

zuhdil:
I'm not sure, the UI is a bit different with the project directory. So, I think we keep the update directory as it is now and create a separate issue if we found a problem with it later.

punchagan:
What exactly do you mean, when you say the UI is a bit different? The Project directory and project update directory seem to be trying to re-using as much code as possible.

punchagan:
Merging based on @zuhdil explanation that it is missing fields that are causing a problem here, not incorrect values in the DB.  The `ProjectUpdateDirectorySerializer` is already doing the right thing.

PR #3536, opened 2019-04-09
Restrict project access to users employed by hierarchy owner and it's collaborator org

punchagan:
This PR adds the new concept of a collaborator organisation. It builds on top of the concept of shadow organisations, which are content owned organisations that are clones of other organisations in RSR. Collaborator organisations are shadow organisations whose content owner and the original organisation are the same organisation. They are shadows for the content owner organisation that are managed by the content owner themselves. 

This PR restricts access to projects, that are part of project hierarchies that have `enable_restrictions` set to True, to employees of the owner organisation of the project hierarchy and it's collaborator organisation. 

The PR acknowledges the fact that the concept of `ProjectHierarchy` and access to partners are being tightly coupled, here. We need the concept of an "owner" for a project, and in case of projects in a hierarchy the owner is the organisation that owns the hierarchy. For other projects, they could be the primary organisation or reporting organisation.  This refactor is delayed for further discussion of another use case to see what works well and what doesn't. 

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry

### DIFF
--- a/akvo/rsr/management/commands/create_collaborator_organisation.py
+++ b/akvo/rsr/management/commands/create_collaborator_organisation.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+
+# Akvo Reporting is covered by the GNU Affero General Public License.
+# See more details in the license.txt file located at the root folder of the Akvo RSR module.
+# For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
+
+"""Create a collaborator organisation for a given organisation.
+
+Usage:
+
+    python manage.py create_collaborator_organisation <org-id>
+
+"""
+
+import sys
+
+from django.core.management.base import BaseCommand
+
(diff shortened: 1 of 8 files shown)


zuhdil:
So, one organisation will only have one "collaborator organisation" for all of the organisation's projects, is that correct?

If the "shadow organisation" will be dropped and replaced with this "collaborator organisation", than I think the `content_owner` and `original` attributes are somewhat redundant, but I'm sure you already aware of it.

punchagan:
The `content_owner` attribute is not redundant, because that is what allows the admins of the content owner organisation to administer this organisation. The `original` attribute may be a misnomer, and may be it should be changed to `collaborator`, but I'm not particularly keen on doing that right now. I'll do that once we decide that we are indeed dropping shadow organisations, and actually remove all references to a shadow organisation.

PR #3542, opened 2019-04-17
Issue/3520 spa environment setup

zuhdil:
* The SPA is served by nginx in `/my-rsr` path.
* Added a new directory `/var/akvo/rsr/sparoot/` in the `rsr-nginx` image to host the SPA distribution.

- [ ] Copyright header
- [ ] Code formatting
- [ ] Documentation
- [ ] Change log entry


scripts/docker/dev/start-django.sh
```diff
@@ -21,6 +21,13 @@ fi
 npm run devw&
 popd
 
+pushd akvo/rsr/spa
+if [[ ! -d "node_modules" ]]; then
+  npm install
```
punchagan: Should this be a `yarn install`? I see a `yarn.lock` file was committed to the repo. I'm guessing we want to do a `yarn install` ? @martinchristov ?

zuhdil: I'm a little reluctant to add yarn to the image because it will add more time for the image build process, using `npm install` here everything can run well and no problems found.

I can update it and use `yarn` if it is a must.

punchagan: If `yarn` is the new standard, we can switch to using `yarn` for both the spa and other static assets. But, having a `yarn.lock` file, and not using `yarn` seems like a smell, to me.

zuhdil: I think `yarn` vs. `npm` is only a matter of preference, but I could be wrong.

I understand your concern. We can do one of the following;
- remove `akvo/rsr/spa/package-lock.json`, add `yarn` binary to the images, and use yarn to manage the SPA dependencies.
- remove `akvo/rsr/spa/yarn.lock`, and continue using `npm` to manage the SPA dependencies.

Please give your suggestion and I will update the PR.

akvo/rsr/spa/build/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Akvo RSR</title>
+    <link href="/my-rsr/styles/style.css" rel="stylesheet">
+</head>
+<body>
+<div id="root"></div>
+<script type="text/javascript" src="/my-rsr/bundle.js"></script>
```
dlebrero: As it is, the `bundle.js` cannot be cached. I guess that we will implement some kind of caching, like adding a hash to the js file name?

docker-compose.override.yaml
```diff
@@ -3,12 +3,18 @@ version: '3'
...
   mainnetwork:
     ports:
-      - "80:8000"
+      - "80:80"
+      - "8080:8080" # for webpack HMR
   web:
     build:
       context: .
       dockerfile: Dockerfile-dev
+  nginx:
```
dlebrero: In Lumen, we use the same WebPack dev server as a proxy to the backend, so that we do not need an nginx server running.

Given that Nginx is quite light, probably it is ok to leave it as it is.

scripts/docker/ci/end-to-end.sh
```diff
@@ -39,16 +39,34 @@ if [[ "${RESPONSE_STATUS}" -ge ${HTTP_ERRORS_START} ]]; then
...
-
 while read -r path; do
   # test asset
   test_http_status "${HTTP_OK}" "$BASE_URL$path"
   # test sourcemap
   test_http_status "${HTTP_OK}" "$BASE_URL$path.map"
 done <<< "$ASSETS"
+
+# SPA endpoint end routing
+SPA_URL="$BASE_URL/my-rsr"
```
dlebrero: Thanks a lot for the tests!

nginx/Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN curl -fsSLO --compressed "https://nodejs.org/dist/v$NODE_VERSION/node-v$NODE
 ENV DJANGO_SECRET_KEY=any
 
 RUN cd akvo/rsr/front-end && npm install && rm -rf node_modules && npm install && npm run prod
+RUN cd akvo/rsr/spa && npm install && rm -rf node_modules && npm install && npm run production
```
dlebrero: Do we really need the same workaround as in line 12?

scripts/docker/dev/start-django.sh
dlebrero: Up to @martinchristov to decide. I personally do not know the difference between them. Probably if we wait for a week, there will be a new "standard" 😆

scripts/docker/dev/start-django.sh
```diff
@@ -21,6 +21,13 @@ fi
 npm run devw&
 popd
 
+pushd akvo/rsr/spa
+if [[ ! -d "node_modules" ]]; then
```
dlebrero: Maybe there is a better way to know if `npm install` must be run?

scripts/docker/dev/start-django.sh
martinchristov: Hey guys. We could drop yarn and just stick to plain npm. I've been using yarn because it used to be faster than npm, but I just checked the latest state of affairs and npm has caught up to speed.

scripts/docker/dev/start-django.sh
martinchristov: Not sure what you mean, but `npm install` does it's own check of whether there's anything missing in "node_modules" and installs only then.

akvo/rsr/spa/build/index.html
martinchristov: Sure we can add a hash prefix in order to prevent caching. Is that what you meant?

dlebrero: Yes. For the MVP, not required, but required for the production launch.

scripts/docker/dev/start-django.sh
dlebrero: Checking if the directory `node_modules` exists before running `npm install` is probably due to `npm install` is/was too slow to run in each and every restart of the container. 

Maybe this is no longer the case, or the delay is acceptable, or the npm install can be run in the background the same way that `npm start` is run in line 27?

martinchristov: `npm install` does take a while because it also checks for updates to the packages. It might actually be good if we can choose when to run an 'npm install' by deleting the node_modules folder when publishing.

martinchristov: Actually I just checked `yarn install` vs `npm install` and the result is 1 second vs 15-60 seconds.

dlebrero: There is the option to use docker layering to run `npm install` just when the package.json has changed, but that means not being able to see the node_modules content from the host machine (so not being able to find see it in your Mac filesystem, which probably means not being able to browse the source code of the dependencies from your IDE).

Or we could implement our own change detection mechanism of the package.json.

martinchristov: In this case I think it would be better if we just use yarn and run `yarn install` every time.

dlebrero: Agree. Didn't see your `yarn` comment before posting my comment.

scripts/docker/dev/start-django.sh
zuhdil: npm it is then 😃

akvo/rsr/spa/build/index.html
zuhdil: Actually this file is targeted by webpack-dev-server in development mode only, for production webpack use `akvo/rsr/spa/app/index.html` and inject the css & js tag in the file content. The js file is already using `[name].[chunkhash].js` but the css is not yet using hash in the file name.

scripts/docker/ci/end-to-end.sh
zuhdil: Your welcome @dlebrero

scripts/docker/dev/start-django.sh
martinchristov: I wrote the above after doing a quick research, but later when I was running tests of how `npm install` (latest version of npm) is performing I discovered that it still takes an unreasonably large amount of time. Even when there is nothing new to install npm takes some 15-20 seconds just to figure this out, whereas yarn takes just 1 second.
Perhaps we should use yarn after all.

akvo/rsr/spa/build/index.html
dlebrero: Awesome! Feel free to do the CSS hashing as a different work item

scripts/docker/dev/start-django.sh
punchagan: We could use `npm` for now, to get this PR merged quicker. And then later move all the `npm` stuff to `yarn`.

akvo/rsr/spa/build/index.html
punchagan: Let's do this as a different work item! @zuhdil

Issue #3565, opened 2019-05-09
Remove unused data models


AkvoAnt:
Focusing on dimensions in results framework

PR #3574, opened 2019-05-15
[#3565] Remove IndicatorPeriod{Actual,Target}Dimension models

punchagan:
The actual dimension models have been replaced by the `IndicatorDimension` and
the `Disaggregation` models. They still need to be extended to be used for the
target dimensions, though.

- [ ] Documentation


punchagan:
[actual-target-dimensions.xlsx](https://github.com/akvo/akvo-rsr/files/3193463/actual-target-dimensions.xlsx)

There is some data, but most of it seems useless. We could inform the partners about deleting this data, if it seems useful.

punchagan:
This data has not been available in the UI for a long time now, and nobody seems to complain. I'm going ahead and merging the deletion of these unused models.

PR #3613, opened 2019-05-29
[#3612] Setup handlers on dynamic vocabulary dropdowns only once

### LINKED ISSUE #3612: Dynamically populated dropdown options work incorrectly
When adding new partials in the same section, options chosen in the dynamically created dropdowns for vocabulary options are removed.

punchagan:
The code that is setting up the change handlers, and creating the dynamic
dropdowns, and switching between them is called each time a new partial is
added. This causes an existing selection in a dropdown to be removed.

This commit fixes that by ensuring that the dynamic dropdown is not setup
multiple times, if it has already been setup.

Closes #3612

  - Choose a project where EUTF is the reporting organisation
  - Ensure that EUTF organisation has codelist data (can be copied from live or test, if not already present in local DB).
  - Try to add a couple of administrative locations peekN6IF2Z

- [ ] Documentation


punchagan:
peekBC6I2Z

zuhdil:
Looks good.

PR #3627, opened 2019-06-05
Remove fake-migration-flag

### LINKED ISSUE #3626: Remove fake-migration-flag
This was added as part of the migration from CloudVPS to ElephantSQL. 

It is not needed anymore and it is causing some issue with the work at #3617, where we 
had to implement a workaround for the fake migrations to not run (https://github.com/akvo/akvo-rsr/blob/9227d343a2209fe7a396373555d9321b3a0d3256/ci/rsrchart/templates/deployment.yaml#L88)
when starting from an empty DB.

dlebrero:
Fixes #3626


punchagan:
Thanks, merged!

Issue #3578, opened 2019-05-17
Including aggregated data for disaggregation values at the parent RF level


AkvoAnt:
We need to enable the aggregated data at a parent project results framwork for disaggregation values; both in My Projects and the Results tab of the public project page.

punchagan:
Closed in #3702

PR #3702, opened 2019-07-19
[#3578] Aggregate disaggregations to parent level

zuhdil:
Also some test cases as feature behaviors contract.

- [ ] Documentation


akvo/rsr/models/result/indicator_period_disaggregation.py
```diff
@@ -0,0 +1,52 @@
...
+from django.db import models
+from akvo.rsr.mixins import TimestampsMixin
+from django.utils.translation import ugettext_lazy as _
+
+
+class IndicatorPeriodDisaggregation(TimestampsMixin, models.Model):
+    """
+    Model to hold aggregated disaggregation updates
+    """
+
```
punchagan: The model needs a `project_relation` attribute.

akvo/rsr/models/result/indicator_period_disaggregation.py
```diff
@@ -0,0 +1,52 @@
...
+    period = models.ForeignKey(
```
punchagan: Would it be worth making an abstract model where we have all these fields? And use that everywhere we need these fields?

punchagan: Yeah, I think that's good enough. May be also use that for the `IndicatorPeriodData` ? This would make the updates, and the disaggregations also be somewhat consistent.

zuhdil: Yes, it can also used in `IndicatorPeriodData`, the mixin's name is still appropriate to use there.

punchagan: Can we just make it `IndicatorValueMixin` or `IndicatorUpdateMixin` ?

PR #3705, opened 2019-07-19
Allow protecting partner sites using a global password

punchagan:
- [ ] Documentation


punchagan:
peek278O5Z

Here's a screenshot showing what the feature looks like

We could potentially add some CSS to make the password form prettier. @martinchristov can you help with this? 

We should also probably change the text a little bit.

PR #3740, opened 2019-08-12
[#3738] period actual comment clean-up

### LINKED ISSUE #3738: Period.actual_comment clean-up.
Ignore period update without comment

zuhdil:
- [ ] Documentation

Fixes #3738

### DIFF
--- a/akvo/rsr/management/commands/period_actual_comment_cleanup.py
+++ b/akvo/rsr/management/commands/period_actual_comment_cleanup.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+
+# Akvo Reporting is covered by the GNU Affero General Public License.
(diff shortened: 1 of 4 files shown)


akvo/rsr/management/commands/period_actual_comment_cleanup.py
```diff
@@ -0,0 +1,19 @@
...
+
+from django.db.models import Q
+from akvo.rsr.models import IndicatorPeriod
+
+
+class Command(BaseCommand):
+    help = "Clean up period.actual_comment that contains only period update dates"
+
+    def handle(self, *args, **options):
+        periods = IndicatorPeriod.objects.filter(~Q(actual_comment=''))
```
punchagan: This could be written with an `exclude` query, instead of `~Q`
We could also filter for periods which have child periods. 

But for a one-time running script, I guess it's not a big deal.

PR #3764, opened 2019-09-05
Support for ReportServer in training envs

dlebrero:
Added ReportServer to the RSR pod.

We always copy the RS production DB, even when starting from an empty RSR DB, as we want to have reports in the training env.

It is possible that RS does not work in an empty RSR due to the DB tables not existing when the RSR readonly user is created. Not checking this until somebody actually bothers to use an empty RSR.

### DIFF
--- a/ci/k8s/deployment.yml
+++ b/ci/k8s/deployment.yml
@@ -23,12 +23,12 @@ spec:
         ports:
         - containerPort: 80
           protocol: TCP
-        - containerPort: 8080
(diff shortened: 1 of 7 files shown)


punchagan:
I've tested this branch with the temporary commit that enabled deploying of containers from this branch. The training server worked well with the reports - I tried a couple of simple reports. 

I'm merging this branch after removing the temporary commits to enable and disable deploying containers built on this branch.

PR #3789, opened 2019-09-19
[#3787] Support nested location creation in OrganisationViewSet

### LINKED ISSUE #3787: Adding organisation via PE
https://trello.com/c/zII0ZO0m/27-adding-organisation-via-pe

punchagan:
When latitude and longitude data is present in the POST data sent to
`/v1/organisation`, the co-ordinates are used to create a location for the
newly created organisation.

Closes #3787

- [ ] Documentation


punchagan:
This needs to be verified against the old PE. I think it may end-up creating two project locations.

PR #3806, opened 2019-09-25
3805 deep linking

martinchristov:
- [ ] Documentation


punchagan:
I pushed a couple of commits to change the link hashes in the IATI page. 

It looks like the scrolling doesn't work when trying to deep link to a period - there seems to be an error in the scrolling code `TypeError: $period is undefinedsection5.jsx:251:12
    Section5 section5.jsx:251`

martinchristov:
Please provide a URL

PR #3811, opened 2019-10-01
WIP: Add project level permissions for users

punchagan:
This is an implementation of project level permissions that allows adding users directly to projects with specific roles.  

This permission backend by-passes the current permission backend, and any organizations that have enabled this feature are going to use only this permission backend, and access to a project is given to a user only when they have been explicitly added to a project. 

The PR is incomplete, in that it doesn't add a UI or the API required to enable this feature, but it implements the core of the authentication backend, and some tests to verify that the implementation is reasonable and actually works.


akvo/rsr/project_role_auth.py
```diff
@@ -0,0 +1,78 @@
...
+    rules = rules.replace('(', '').replace(')', '').replace('|', '').split()
+
+    rule_to_group_mapping = {
+        'is_org_admin': [GROUP_NAME_ADMINS],
+        'is_org_user_manager': [GROUP_NAME_USER_MANAGERS],
+        'is_org_me_manager_or_project_editor': [GROUP_NAME_PROJECT_EDITORS, GROUP_NAME_ME_MANAGERS],
+        'is_org_me_manager': [GROUP_NAME_ME_MANAGERS],
+        'is_org_user': [GROUP_NAME_USERS],
+        'is_org_enumerator': [GROUP_NAME_ENUMERATORS],
+    }
```
zuhdil: I'm not really sure about this, it's a kind of duplication with predicate definitions, _i.e._ we need to maintain it in two different places.

Is it possible if the project role permission logics are implemented as `rules.predicate` instead of auth backend? I think that will be more future-proof and can eliminate the problem of parsing predicate names and  assumes only `|` operator will be used.

And I've just noticed that we are using a very old version of `django-rules`. but I don't think that's an important issue right now.

punchagan: You are correct about the duplications, but we could get rid of the duplications and maintain the same list of groups to permission mapping that is used both by the project permissions layer and the organisation employment permissions layer. 

I see that there would be some gains to use a `rules.predicate` for the project level permissions, but I don't really like the fact that each predicate is run separately for each rule, and the org level rules are sometimes quite slow to compute. Also, we'd need to make sure we correctly add project level predicates wherever we have org level predicates, and make sure we keep doing that in the future too. 

The other option could be to combine the project level checks into the org level predicates, and call them something more generic. But, still this would mean the predicates are called individually, and each of the checks is run multiple times inside each predicate. 

Having the project level permissions as a separate layer in front of the organisation permissions layer allows us to be able to just run through one function and decide if a user has any project level permissions or not.

punchagan:
Please don't merge this until we have a UI for the permissions, since one of the partners is using the old permissions system, and it would be nice to not have them locked up without a UI.

PR #3812, opened 2019-10-01
[#3774] Add command to reverse geocode lat,lng to city names

### LINKED ISSUE #3774: S7 location issues
- [x] blank screen for project 7595
- [x] unhandled encoding at administratives > codes
- [ ] need to handle better missing "city" field on pre-existing projects

punchagan:
Closes #3774

- I've run this locally to verify that it works for a few hundred locations, but when I say works - we get back some city names for some lat/lng values. 

- How do we verify that what we are back-filling is actually correct data? 

- What happens if some of the data that we back-filled is actually incorrect? How does this effect the partners? Since we can't really identify incorrect data, as of now, rolling back these incorrect back-fills can only happen on a case-by-case by the partners.

### DIFF
--- a/akvo/rsr/management/commands/reverse_geocode_city_names.py
+++ b/akvo/rsr/management/commands/reverse_geocode_city_names.py
@@ -1,0 +1,124 @@
+# -*- coding: utf-8 -*-
+
+# Akvo Reporting is covered by the GNU Affero General Public License.
+# See more details in the license.txt file located at the root folder of the Akvo RSR module.
+# For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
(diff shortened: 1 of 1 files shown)


zuhdil:
The frontend has moved to the Google API, what do you think about using it here too, for consistency?

punchagan:
We could do that, but it seems like a lot of work for a one-time script. :)
I'll wait to hear back from our partners after looking at the sample data, before doing anything with this.

punchagan:
We got a go-ahead on this one. @zuhdil should we run the script the way it is, or switch to Google API? We got one of the partners to verify that their project locations are correct with the API that we are using...

Issue #3814, opened 2019-10-02
Jump to results and indicators from the preview


martinchristov:

martinchristov:
![](https://github.trello.services/images/mini-trello-icon.png) [Jumping to results and indicators from the preview](https://trello.com/c/s8zxGkhX/22-jumping-to-results-and-indicators-from-the-preview)

PR #3815, opened 2019-10-02
[#3814] s5 jump to results & indicators

martinchristov:
- [ ] Documentation

### DIFF
--- a/akvo/rsr/spa/app/modules/editor/section5/section5.jsx
+++ b/akvo/rsr/spa/app/modules/editor/section5/section5.jsx
@@ -16,6 +16,7 @@ import './styles.scss'
 import Accordion from '../../../utils/accordion'
 import Indicators from './indicators'
 import AutoSave from '../../../utils/auto-save'
+import { useForceUpdate } from '../../../utils/hooks'
 import {addSetItem, removeSetItem, fetchSetItems} from '../actions'
 import api from '../../../utils/api'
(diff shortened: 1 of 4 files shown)


martinchristov:
1. What text/icon should the button have? Currently it's "Open" not sure if this is best.
2. There is a slight delay after clicking the button to give time to the modal to close & accordion to switch before scroll is performed. Do you feel it's too much the delay?

punchagan:
> There is a slight delay after clicking the button to give time to the modal to close & accordion to switch before scroll is performed. Do you feel it's too much the delay?

The first time it seemed like a little too long. But, after I used it once, it didn't feel like too long.

punchagan:
> What text/icon should the button have? Currently it's "Open" not sure if this is best.

Should it be a separate button? Or could the titles themselves be clickable?

punchagan:
We should probably move this discussion to Zulip to get inputs from Anthony, as well, I guess.

martinchristov:
We'll have to merge to develop for Anthony to see it.
As for titles being clickable - they already are clickable but to expand-collapse. I think it should remain as so and the "jump to" as a separate button as it is not the primary action.

punchagan:
> What text/icon should the button have? Currently it's "Open" not sure if this is best.

We could use the word `Edit` instead of `Open`.  Open could also refer to opening the folded accordion.

PR #3817, opened 2019-10-03
[#3813] Add properties and filters to /editable_project.

### LINKED ISSUE #3813: Add properties and filters to /editable_project
The new /projects page shows a bit more information than currently available at the endpoint.
Please add properties:

 - [x] privacy
- [x] sectors[]
- [x] parent { name, id }

Please add filters:

- [x] sector
- [x] country

Screenshot 2019-10-02 at 9 44 54

zuhdil:
- [ ] Documentation

Closes #3813


punchagan:
I've added a couple of commits on top of this to clarify a couple of things.  The `editable_project` end-point was actually returning all the projects that the user can perform an action on, based on their role, not just projects they can edit. Users who cannot edit projects should be sent to the `/myrsr/my_project/<id>` URL, instead of taking them to the editor.  I have changed the front-end code to do this. 

Also, I force pushed to merge Zuhdil's two commits, since the second commit was just minor fixes on top of the first one.

punchagan:
To make it easier to work with rebased branches, I've added a helper script `scripts/devhelpers/fetch-pull-request`

PR #3819, opened 2019-10-03
Issue/3816 calc totals

martinchristov:
- [ ] Documentation

### DIFF
--- a/akvo/rsr/spa/app/modules/editor/section7/recipient-regions/recipient-regions.jsx
+++ b/akvo/rsr/spa/app/modules/editor/section7/recipient-regions/recipient-regions.jsx
@@ -4,6 +4,7 @@ import { useTranslation } from 'react-i18next'
 
 import FinalField from '../../../../utils/final-field'
 import ItemArray from '../../../../utils/item-array'
+import CalcTotalChecker from '../../../../utils/calc-total-checker'
 import InputLabel from '../../../../utils/input-label'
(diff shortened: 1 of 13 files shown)


akvo/rsr/spa/app/modules/editor/section7/recipient-regions/recipient-regions.jsx
```diff
@@ -114,6 +115,7 @@ const RecipientRegions = ({ formPush, validations, showRequired, errors }) => {
           </Button>
         )}
       />
+      <CalcTotalChecker section="7" path="recipientRegions" prop="percentage" />
```
punchagan: If there are recipient regions and countries both specified, the combined total should add up to 100%, not separately. 

See this [Zulip thread](https://akvo.zulipchat.com/#narrow/stream/131441-rsr/topic/Edit.20Error.20message.20IATI.20publication.20country.20percentage/near/173780667), for more information.

punchagan:
image

May be we should move the recipient country and regions to be together? The remaining %age to be filled is not visible when even one of the the locations is expanded. 

Looks good, otherwise!

martinchristov:
Ooh... I hadn't noticed this because the order I was looking at was:

1. Location items
2. Recipient countries
3. Recipient regions

However, there is a conditional in the code which puts Recipient countries on top for EUTF only.
Let's bring @AkvoAnt into this

punchagan:
@martinchristov Do we still want to keep this open? Or just close this as stale?

PR #3851, opened 2019-10-29
Issue/3850 child projects

martinchristov:
- [ ] Documentation


punchagan:
It looks like changing a related project doesn't work correctly and raises a 405. The URL seems to be incorrect. The url being hit with the PATCH request right now is `/rest/v1/related_project?relation=1<id>/&format=json`. It should be `/rest/v1/related_project/<id>&format=json`.

martinchristov:
Those should be fixed

martinchristov:
Thanks for catching those

PR #3877, opened 2019-11-26
[#3876] Python generated reports pipeline

zuhdil:
- [ ] Documentation


punchagan:
This looks good to me, in general. Would it be possible to have a simple PDF/excel report also, in this commit, so that it'll be easy to test once we merge this?

punchagan:
Yes, I was also concerned about the duplication, and this makes sense to me!

akvo/rsr/views/py_reports.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+import os
+import pdfkit
+from django.http import HttpResponse
+from django.template.loader import render_to_string
+
+
+def check(request):
+    is_reports_container = os.getenv('DJANGO_PORT', None) == '9000'
```
dlebrero: Use a specific env variable "IS_REPORT_CONTAINER". 

It is going to be very confusing if we change the port number and then suddenly reports stop working.

docker-compose.yaml
```diff
@@ -15,6 +15,13 @@ services:
     network_mode: service:mainnetwork
     environment:
       - DJANGO_SECRET_KEY=secretkey
+  reports:
+    environment:
+      - DJANGO_SECRET_KEY=secretkey
+      - SKIP_FRONTEND_BUILD=yes
```
dlebrero: I would opt to use one env variable "IS_REPORT_CONTAINER" for all, instead of 3 different skips, but either is ok.

docker-compose.ci.yaml
```diff
@@ -5,5 +5,8 @@ services:
   web:
     image: rsr-backend:dev
     command: "true"
+  reports:
```
dlebrero: I think we do not use this container for the CI server. Maybe we do not need to start it at all?

punchagan:
`ci/k8s/deployment.yml` and `ci/training-envs/templates/deployment.yaml` still need to be updated to create the containers on k8s. @zuhdil would you be taking care of this?

PR #3890, opened 2019-12-04
[#3876] Update k8s deployment file to create reports container

punchagan:
I'm not sure if this is the right way of doing things, but I created a WIP PR, so we can have a chat about this. 

- [ ] Documentation

### DIFF
--- a/ci/k8s/deployment.yml
+++ b/ci/k8s/deployment.yml
@@ -121,6 +121,69 @@ spec:
               key: django-secret-key
         - name: ENABLE_STATS
           value: "yes please"
+      - name: rsr-reports
+        image: eu.gcr.io/akvo-lumen/rsr-backend:${TRAVIS_COMMIT}
+        imagePullPolicy: Always
+        resources:
+          requests:
+            cpu: "50m"
+          limits:
+            cpu: "4000m"
+        ports:
+        - containerPort: 9000
+          protocol: TCP
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 9000
+          initialDelaySeconds: 30
+          timeoutSeconds: 1
+          failureThreshold: 12
+          periodSeconds: 10
+        readinessProbe:
+          httpGet:
+            path: /healthz
+            port: 9000
+          initialDelaySeconds: 30
+          timeoutSeconds: 1
+          failureThreshold: 12
+          periodSeconds: 10
+        volumeMounts:
+            - name: "secret-config"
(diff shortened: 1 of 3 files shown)


ci/k8s/deployment.yml
```diff
@@ -121,6 +121,65 @@ spec:
...
+          valueFrom:
+            secretKeyRef:
+              name: "rsr-secret"
+              key: gunicorn-debug
+        - name: DJANGO_SECRET_KEY
+          valueFrom:
+            secretKeyRef:
+              name: "rsr-secret"
+              key: django-secret-key
+        - name: ENABLE_STATS
```
dlebrero: I am missing the env var that tells the container to behave like a report server (same for the other yaml file)

dlebrero: I am missing the env var that tells the container what port to use (same for the other yaml file).

See the docker-compose.override.yaml file.

ci/k8s/deployment.yml
```diff
@@ -121,6 +121,65 @@ spec:
               key: django-secret-key
         - name: ENABLE_STATS
           value: "yes please"
+      - name: rsr-reports
```
dlebrero: YAML is crazy. Maybe you want to remove some of the duplication using references https://blog.daemonl.com/2016/02/yaml.html

punchagan: woah! That is indeed crazy. I removed some of the duplication using references.

ci/k8s/deployment.yml
punchagan: Fixed, thanks

punchagan: Fixed, thanks

ci/k8s/deployment.yml
dlebrero: Where you able to test that the new YAML is correct?

punchagan: I used this to verify the yaml. 
`python -c 'import sys, yaml, json; json.dump(yaml.load(sys.stdin), sys.stdout, indent=4)' < ci/k8s/deployment.yml | less`

The training yaml was tricky because it is a template. I just ended up deleting all the template tags and verifying

dlebrero: Very clever!

The training yaml can be tested by deploying a new training env. Do you mind doing that?

punchagan: Sure, I can do that!

punchagan: I've tested this, and deployed it on `rsr4`. It seems to work!

PR #3892, opened 2019-12-05
[#3876] Custom reports for EUTF Syria.

zuhdil:
- [ ] Documentation

### DIFF
--- a/akvo/rsr/staticmap.py
+++ b/akvo/rsr/staticmap.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+
+"""Akvo RSR is covered by the GNU Affero General Public License.
+
+See more details in the license.txt file located at the root folder of the Akvo RSR module.
+For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
+"""
+
+import collections
+from django.conf import settings
+
+
+Coordinate = collections.namedtuple('Coordinate', ['latitude', 'longitude'])
+
+Size = collections.namedtuple('Size', ['width', 'height'])
+
+
+class MapquestAdapter(object):
+
+    URL = "https://www.mapquestapi.com/staticmap/v5/map?key={}&locations={}"
+
+    def __init__(self, key):
+        self.key = key
+
+    def get_url(self, locations, size=None, zoom=None):
+        location_groups = self._group_location(locations)
+        location_strings = [
+            '{},{}|marker-{}'.format(c.latitude, c.longitude, v)
+            for c, v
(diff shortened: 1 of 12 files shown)


akvo/rsr/staticmap.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+Coordinate = collections.namedtuple('Coordinate', ['latitude', 'longitude'])
+
+Size = collections.namedtuple('Size', ['width', 'height'])
+
+
+class MapquestAdapter(object):
+
+    URL = "https://www.mapquestapi.com/staticmap/v5/map?key={}&locations={}&defaultMarker=via-lg-1890ff-6b72a8&zoom=8"
```
punchagan: When there are multiple lat/lng, hardcoding the zoom level doesn't seem to work correctly. For example, with these two co-ordinates `28.704059,77.10249||19.387389,85.051541` setting the zoom level to 8 makes both the points disappear from the map, and we zoom in at the mid-point of these two points. Instead not setting the zoom value or setting it to 6 makes the map appear correctly. 

I think the zoom level here needs to be dynamically figured out based on the number of locations and their distance from each other, or may be even just the number of points. The API doesn't seem to have a max-zoom, so if there is a single coordinate, it seems to be too zoomed in, in which case we could explicitly set a zoom level to 8.  

We could have some kind of a heuristic to calculate the dimensions of the bounding box, and set the zoom level accordingly.

akvo/rsr/views/py_reports.py
```diff
@@ -31,3 +40,222 @@ def check(request):
...
+    print(request.GET, show_comment, start_date, end_date)
+
+    if request.GET.get('show-html', ''):
+        return HttpResponse(html)
+
+    now = datetime.today()
+
+    return _make_pdf_response(
+        html,
+        options={
```
punchagan: We only have a couple of reports, so this may be fine. But, if most of this configuration seems pretty standard to me, and may be we can make it a default that we over ride.

akvo/templates/reports/grids.css
```diff
@@ -0,0 +1,38 @@
+
```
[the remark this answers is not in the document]
zuhdil: I'm agree with you. But because the `wkhtmltopdf` is not running under a web server, it can only do an external request to load resources (css, images, etc.). it's not able to resolve relative url (url without the hostname). 

I can't seem to find an elegant way to load the CSS like any other CSS are loaded in Django templates. Beside that, I think it's unnecessary to do an external request to load the internal CSS. That's why I'm using the Django template's `include` tag to embed the CSS, and separate the CSS file just for modularity purpose.

PR #3902, opened 2019-12-27
Port the slow REST API migration tests to proper tests

punchagan:
This PR gets rid of the slow migration tests, that were written when changing DRF versions. 
- The tests are hard to maintain, and haven't really gotten too many updates. 
- This PR ports all the tests which are testing the currently used API endpoints to normal tests, alongside the other existing tests. 
- The `expected_responses.json` file often creates problems for new PRs, causing conflicts and needing PRs to be fixed, even if the two PRs are independent changes. 
- This change cuts down the time to run builds by about 4 to 5 minutes, which is a great side effect of this change!
- This PR will also make the porting of the code-base to Python3, a little easier, since the old slow tests are tedious to run and update.


punchagan:
@zuhdil This PR is ready for review. Please take a look, when you get some time.

PR #3903, opened 2019-12-28
Move to Python3!

punchagan:
This pull request has commits on top of the #3902 PR that ports the slow migration tests to proper tests. Reviewing this would be easier if #3902 is merged first. 

The PR contains the changes for moving the code base to Python3. The changes here are no longer compatible with Python2.  We should only merge this PR once we are ready to move completely to Python3.


punchagan:
@zuhdil This PR is ready for review. Please take a look, when you get some time. It would be easier to review if #3902 is reviewed and merged, before taking a look at this.

PR #3946, opened 2020-02-05
Issue/3876 weasyprint

zuhdil:
- [ ] Documentation

The main reason for using WeasyPrint is because of better support for page breaking behaviour, with added bonus on easier to create TOC page.

And better support for CSS standards!

### DIFF
--- a/akvo/rsr/views/py_reports.py
+++ b/akvo/rsr/views/py_reports.py
@@ -20,6 +20,8 @@ from django.db.models import Q
 from django.http import HttpResponse, HttpResponseBadRequest
 from django.shortcuts import get_object_or_404
 from django.template.loader import render_to_string
+from weasyprint import HTML
+from weasyprint.fonts import FontConfiguration
 
 
 DEFAULT_PDF_OPTIONS = {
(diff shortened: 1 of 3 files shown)


akvo/rsr/views/py_reports.py
```diff
@@ -72,8 +74,10 @@ def render_organisation_projects_results_indicators_map_overview(request, org_id
         in projects
     ]
 
+    now = datetime.today()
+
     html = render_to_string(
-        'reports/organisation-projects-results-indicators-map-overview.html',
+        'reports/wp-organisation-projects-results-indicators-map-overview.html',
```
punchagan: I'd recommend deleting the other template, if we are not going to use it, and replacing it with the new template. 

Is it important to prefix these templates with `wp-` are they different from the other template that we write for `wkhtmltopdf`?  

Are we going to switch to using WeasyPrint for every report? Or do we still want to keep using `wkhtmltopdf`?

punchagan: These changes can come in a separate PR, though, if that works better.

PR #3952, opened 2020-02-10
[#3876] Cleanup and use Weasyprint for project report.

zuhdil:
- [ ] Documentation


punchagan:
Could you also remove the `wkhtmltopdf` package if we don't need it any more?

PR #4015, opened 2020-02-27
[#4014] python excel report

zuhdil:
Implements the "Project results indicators simple table" report.

- [ ] Documentation

### DIFF
--- a/akvo/rsr/views/py_reports/project_results_indicators_excel_report.py
+++ b/akvo/rsr/views/py_reports/project_results_indicators_excel_report.py
@@ -1,0 +1,185 @@
+# -*- coding: utf-8 -*-
+
+"""Akvo RSR is covered by the GNU Affero General Public License.
+
+See more details in the license.txt file located at the root folder of the
+Akvo RSR module. For additional details on the GNU license please
+see < http://www.gnu.org/licenses/agpl.html >.
+"""
+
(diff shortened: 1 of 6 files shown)


akvo/rsr/views/py_reports/project_results_indicators_excel_report.py
```diff
@@ -0,0 +1,183 @@
...
+                ws.set_cell_value(row, 3, indicator.baseline_year)
+                ws.set_cell_style(row, 4, Style(alignment=Alignment(horizontal='right')))
+                ws.set_cell_value(row, 4, indicator.baseline_value)
+                ws.set_cell_style(row, 5, Style(alignment=Alignment(wrap_text=True)))
+                ws.set_cell_value(row, 5, indicator.baseline_comment)
+
+                for period in indicator.periods.all():
+                    ws.set_cell_value(
+                        row, 6,
+                        utils.get_period_start(period, in_eutf_hierarchy).strftime('%Y-%m-%d'))
```
punchagan: There can be projects where the period dates have not been set correctly, and this would cause a server error. May be handle `start_date` or `end_date` being `None`?

zuhdil: Good catch! Thanks @punchagan

Issue #4072, opened 2020-03-23
Move Memcached out of pod


dlebrero:
Right now we have the Memcached container in the same pod as RSR which means that when RSR is deployed, the cached objects in memcached are lost, making the first bunch or request to RSR specially slow.

Move memcache out of the RSR pod so it survives RSR deploys, unless Memcached itself is changed.

PR #4073, opened 2020-03-23
Move Memcached out of RSR pod

dlebrero:
* Creating a deployment and service for it.
* Config changes for RSR to use the new service.
* Rename memcached container to rsr-memcached so that the config is same as in production.
* Upgrade memcached related containers to latest.

See #4072


punchagan:
Seems like a reasonable change to me. Thanks for working on this, @dlebrero !

PR #4074, opened 2020-03-23
Issue/4069 qualitative indicators

martinchristov:
Test one with program: 7350
http://localhost/my-rsr/programs/7350
Look under "Narrative Report" result

You would need to add some sample data to properly test.
Add an update to project "Project EIS: End Impunity of SGBV" Indicator "0. External factors and risks".

Then add a sub-project to the one above and post a text update to that.


loicsans:
From what I can see, it's fine

Issue #3896, opened 2019-12-16
Program overview


martinchristov:
### Endpoints

**/program/:id/results**
```
  [
    {
      id,
      title: 'EUTF result 01',
      indicators: [
        {
          id,
          title: 'Number of migrants in transit forcibly...',
          description: '...',
          periodCount: 2,
          type: 'quantitative'
        }
      ]
    }
  ]
```

**/program/:id/indicator/:id/**
```
  [
    {
      id,
      periodStart: '',
      periodEnd: '',
      aggregatedValue: 62023,
      aggregatedTarget: 200000,
      disaggregations: [
        { value: 120, title: 'women'}, { value: 201, title: 'children'}, { value: 419, title: 'men' }
      ],
      countries: [{ isoCode: 'ID' }, {isoCode: 'IN'}],
      projects: [
        {
          title,
          countries: [{ isoCode: 'ID' }],
          value: 43
        }
      ]
    }
  ]
```

PR #4076, opened 2020-03-23
[#3896] Aggregates percentage values

zuhdil:
- [ ] Documentation


punchagan:
@zuhdil Thanks for working on this. I've pushed a couple of tests on this, with deliberately failing cases. The tests were written for me to try to understand better what we are doing here. And I am trying to understand the relation between `actual` and `aggregated`. As I see it, aggregated only contains the aggregations from childrens, without any contributions from that project itself. This seems to be the case with `unit` indicators, but not `percentage` indicators. Could you see if my understanding is correct, and if so make that more consistent?  Thanks!

PR #4079, opened 2020-03-25
[#4050] Program reports.

zuhdil:
- [ ] Documentation

Program report is a report that has organisation parameter with `program=true` in the url query param


akvo/rest/views/report.py
```diff
@@ -52,6 +52,45 @@ def report_formats(request):
     })
 
 
+@api_view(['GET'])
+def organisation_reports(request):
+    queryset = Report.objects.prefetch_related('organisations', 'formats')\
+        .filter(url__icontains='{organisation}')\
+        .exclude(url__icontains='program=true')
```
punchagan, line 59: Why do we need this kind of a URL? Why can't we have `{program}` for program reports?

akvo/rest/urls.py
```diff
@@ -283,9 +283,15 @@
     url(r'v1/typeaheads/keywords$',
         views.typeahead_keyword,
         name='keyword_typeahead'),
+    url(r'v1/organisation_reports/$',
```
punchagan, line 286: I like this being split out! :+1:

akvo/rest/views/report.py
```diff
@@ -52,6 +52,45 @@ def report_formats(request):
...
+        queryset = queryset.filter(
+            Q(organisations=None) | Q(organisations__in=approved_orgs)
+        ).distinct()
+
+    serializer = ReportSerializer(queryset.distinct(), many=True)
+    return Response({'results': serializer.data})
+
+
+@api_view(['GET'])
+def program_reports(request, program_pk):
```
punchagan, line 75: I think we should add some tests for this end-point as well?

akvo/rest/views/report.py
```diff
@@ -52,6 +52,45 @@ def report_formats(request):
     })
 
 
+@api_view(['GET'])
+def organisation_reports(request):
```
punchagan, line 56: I think we should add some tests for this new end-point, as well. I think we have some tests for the reports end-point that can be moved over or translated.

akvo/rsr/spa/app/modules/reports/reports.jsx
```diff
@@ -15,6 +15,8 @@ const Reports = ({programId, userRdr}) => {
     }
   }, [userRdr])
   const orgs = userRdr && userRdr.organisations ? userRdr.organisations : []
+  const employment = currentOrg && userRdr.approvedEmployments.find(it => it.organisation === currentOrg)
```
punchagan: Is there some way we can move this to the backend? Just wondering. Ideally, I'd want to keep the permissions and roles checks on the backend, to keep them all in one place.

akvo/rest/views/report.py
zuhdil: It's just for convenience, at least for now, because previously the reports are an organisation report, and it won't be easy to update the existing reports that still lives in ReportServer. But once we have migrated all the reports to be generated by python then it will be more doable.

punchagan: How many such reports do we have, which were previously organisation reports and now will be program reports?

zuhdil: Currently 8 reports.

Issue #4083, opened 2020-03-27
Add a "hierarchy" tab to project editor


martinchristov:
This will enable flawless navigation from Program / Hierarchy -> Edit/Create project -> (back to) Program / Hierarchy

To achieve this the FE needs to be provided the lead/program at the `/project/` endpoint (another way could also be suggested).
Currently the parent project is accessible via the `/related_project/` endpoint, but there is no way of knowing whether the parent is the lead/root/program. Providing this as a prop (isProgram) there can also work.

PR #4085, opened 2020-03-27
[#4083] Project part of a program

zuhdil:
- [ ] Documentation

### DIFF
--- a/akvo/rsr/models/project.py
+++ b/akvo/rsr/models/project.py
@@ -1051,6 +1051,15 @@ class Project(TimestampsMixin, models.Model):
         except ProjectHierarchy.DoesNotExist:
             return None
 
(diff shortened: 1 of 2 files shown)


akvo/rsr/models/project.py
```diff
@@ -1051,6 +1051,15 @@ def get_hierarchy_organisation(self):
         except ProjectHierarchy.DoesNotExist:
             return None
 
+    def get_program(self):
+        """Return the program which this project includes."""
+        from akvo.rsr.models import ProjectHierarchy
+        ancestor = self.ancestor()
+        if ProjectHierarchy.objects.filter(root_project=ancestor).count() > 0:
```
punchagan: `.exist()` does the boolean check we want to do here.

Issue #4121, opened 2020-04-20
Disaggregations issues


martinchristov:
- [x] remove unique name restriction
- [x] [-Remove] button doesn't update UI

- [x] full width of S5 for smaller res

PR #4122, opened 2020-04-20
[#4121] Remove disaggregation category unique name restriction.

zuhdil:
- [ ] Documentation


punchagan:
Could we add some tests for this to show that the aggregation of disaggregations in such cases works correctly? We probably want to do some UI changes to accommodate for this, for example in the program overview for example. There could be other places where we need fixes in the UI. But, we could fix in future pull-requests, if that feels more appropriate to do.

Issue #4136, opened 2020-04-23
Handle making media URLs absolute correctly 


punchagan:
Different storage backends seem to provide URLs slightly differently. Google storage gives absolute URLs, as it should! And the default file storage gives relative URLs. We should handle these URLs correctly, wherever we are using them.

PR #4138, opened 2020-04-23
Fix how we create absolute URLs for media

punchagan:
See #4136 

- [ ] Documentation

### DIFF
--- a/akvo/iati/exports/elements/document_link.py
+++ b/akvo/iati/exports/elements/document_link.py
@@ -5,6 +5,7 @@
 # For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
 
 from lxml import etree
+from urllib.parse import urljoin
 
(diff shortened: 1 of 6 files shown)


akvo/iati/exports/elements/document_link.py
```diff
@@ -18,9 +19,11 @@ def document_link(project):
     """
     document_link_elements = []
 
+    BASE_URL = "http://rsr.akvo.org"
```
dlebrero, line 22: Why is the URL hardcoded here but not in the other document_link at akvo/iati/exports/org_elements/document_link.py?

punchagan: The code in both the contexts is slightly different. One of them has the request being passed as a parameter, but not in the other. Ideally, we should change all the elements to get the request as a parameter. I'll send that change in a separate PR.

PR #4147, opened 2020-04-28
Add scripts to fetch and load dumps from test environment

punchagan:
- [ ] Documentation

### DIFF
--- a/akvo/rsr/eutf_signals.py
+++ b/akvo/rsr/eutf_signals.py
@@ -17,6 +17,10 @@ logger = logging.getLogger('akvo.rsr.eutf_signals')
 def set_reporting_organisation(sender, **kwargs):
     """Set the reporting organisation to EUTF organisation."""
 
+    # Disable signal handler when loading fixtures
+    if kwargs.get('raw', False):
+        return
+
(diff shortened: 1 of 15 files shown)


akvo/rsr/eutf_signals.py
```diff
@@ -17,6 +17,10 @@
 def set_reporting_organisation(sender, **kwargs):
     """Set the reporting organisation to EUTF organisation."""
 
+    # Disable signal handler when loading fixtures
```
dlebrero, line 20: Could you explain why this?

punchagan: `loaddata` calls the pre and post save hooks on the model. But, since the related models may not be loaded, the signal handler may fail. The docs recommend this way of disabling the handlers. Ideally, we only need the pre/post save handlers to have this code, but it's just easier to have this everywhere (and may be even make it a re-usable decorator), instead of forgetting to add it in some cases, etc.

https://docs.djangoproject.com/en/1.11/ref/django-admin/#loaddata

punchagan: I will add this note to the commit message, also. Thanks for your question.

Issue #4179, opened 2020-05-12
IATI reports are not stored in Google Storage


dlebrero:
They are still generated in the k8s persistent disk

PR #4180, opened 2020-05-13
[#4179] Use default_storage to create IATI xml files

punchagan:
- [ ] Documentation

### DIFF
--- a/akvo/iati/exports/iati_export.py
+++ b/akvo/iati/exports/iati_export.py
@@ -12,6 +12,8 @@ import os
 from datetime import datetime
 from lxml import etree
 
+from django.core.files.storage import default_storage, FileSystemStorage
+
 ELEMENTS = [
     'iati_identifier',
     'reporting_org',
@@ -56,6 +58,18 @@ ELEMENTS = [
 ]
 
 
+def save_iati_xml(dir_path, filename, items):
+    if isinstance(default_storage, FileSystemStorage):
+        # GoogleCloudStorage doesn't need parent directories to exist
+        os.makedirs(default_storage.path(dir_path), exist_ok=True)
+
+    file_path = os.path.join(dir_path, filename)
+    with default_storage.open(file_path, "wb") as f:
+        f.write(etree.tostring(items, pretty_print=True))
+
+    return file_path
+
+
 class IatiXML(object):
     def save_file(self, org_id, filename):
         """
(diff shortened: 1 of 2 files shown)


punchagan:
Yes, as part of the cron job.

punchagan:
`./manage.py iati_export` to trigger it manually.

dlebrero:
In the [console](https://console.cloud.google.com/logs/viewer?project=akvo-lumen&minLogLevel=0&expandAll=false&timestamp=2020-05-13T08:11:26.448000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2020-05-13T07:11:57.096Z&dateRangeEnd=2020-05-13T08:11:57.096Z&interval=PT1H&resource=k8s_container%2Fcluster_name%2Fproduction%2Fnamespace_name%2Fdefault%2Fcontainer_name%2Frsr-backend&scrollTimestamp=2020-05-13T08:11:04.074856122Z&filters=text:iati_import), I see that the cron jobs do not pick up the correct log level (I expect the log level to be the same as the gunicorn one, that is INFO). 

This makes me wonder if the cronjobs will read the correct config regarding the google storage or not.

punchagan:
Ah, interesting. I'll investigate. :)

dlebrero:
Does https://github.com/akvo/akvo-rsr/blob/1b553b76f15ad72c7b27552def00ebf16f6e9d3d/akvo/iati/exports/iati_org_export.py#L35 need fixing also?

punchagan:
Ah, yes. I did a search for mediaroot, but somehow overlooked this. Thanks for the catch!

punchagan:
> In the [console](https://console.cloud.google.com/logs/viewer?project=akvo-lumen&minLogLevel=0&expandAll=false&timestamp=2020-05-13T08:11:26.448000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2020-05-13T07:11:57.096Z&dateRangeEnd=2020-05-13T08:11:57.096Z&interval=PT1H&resource=k8s_container%2Fcluster_name%2Fproduction%2Fnamespace_name%2Fdefault%2Fcontainer_name%2Frsr-backend&scrollTimestamp=2020-05-13T08:11:04.074856122Z&filters=text:iati_import), I see that the cron jobs do not pick up the correct log level (I expect the log level to be the same as the gunicorn one, that is INFO).

I think it should. We write the env vars to `/etc/environment`. Our [Django settings](https://github.com/akvo/akvo-rsr/blob/develop/akvo/settings/90-finish.conf#L95-L99) have a debug level set to `DEBUG` by default, and the gunicorn debug level is set as `GUNICORN_DEBUG_ARGS`. This environment variable wouldn't be read by the cron jobs, and that's why the cron jobs seem to use the `DEBUG` level instead of `INFO`.

punchagan:
> This makes me wonder if the cronjobs will read the correct config regarding the google storage or not.

It should be easy to test this on the test server. :)

dlebrero:
Merging so you can test in the test server

dlebrero:
By the way, it would be great to fix the debug level. It annoys me when looking at the RSR logs :P

punchagan:
> Merging so you can test in the test server

That seems to work. :)

> By the way, it would be great to fix the debug level. It annoys me when looking at the RSR logs :P

Yes, I'll send a patch to fix that, separately.

PR #4194, opened 2020-05-20
Use a DB cache for project directory entries. 

punchagan:
**NOTE**: This PR needs to be merged and deployed in two stages. I can do the actual merges and deploys, if I can get a review on this. 

We need to run `python manage.py createcachetable` before we can start using the cache. (I don't know if I should add that to the deploy script itself.) The first few commits refactor some code, and configure this cache. The last couple of commits switch over the code to start using this, and add a management command to populate the cache.  

We use a DB cache so that restarts of the container itself, don't reset the cache. The cache data takes about a couple of minutes to populate now.

### DIFF
--- a/akvo/cache.py
+++ b/akvo/cache.py
@@ -1,11 +1,11 @@
 from functools import wraps
 
 from django.conf import settings
-from django.core.cache import cache
+from django.core.cache import cache, caches
 from django.utils.cache import get_cache_key, _generate_cache_header_key
 
 
-def cache_with_key(keyfunc, timeout=settings.RSR_CACHE_SECONDS):
+def cache_with_key(keyfunc, timeout=settings.RSR_CACHE_SECONDS, cache_name='default'):
     """Decorator which applies Django caching to a function.
 
        Decorator argument is a function which computes a cache key
@@ -17,6 +17,7 @@ def cache_with_key(keyfunc, timeout=settings.RSR_CACHE_SECONDS):
         @wraps(func)
         def func_with_caching(*args, **kwargs):
             key = keyfunc(*args, **kwargs)
+            cache = caches[cache_name]
             val = cache.get(key)
             # Values are singleton tuples so that we can distinguish
             # a result of None from a missing key.
@@ -30,7 +31,8 @@ def cache_with_key(keyfunc, timeout=settings.RSR_CACHE_SECONDS):
     return decorator
 
 
-def delete_cache_data(key):
+def delete_cache_data(key, cache_name='default'):
+    cache = caches[cache_name]
     cache.delete(key)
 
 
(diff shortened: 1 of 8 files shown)


punchagan:
Update: I've actually added the part about running the management command to the `start-django.sh` script itself. If that makes sense, we can merge this PR in a single commit.

dlebrero:
Few questions to understand the problem+solution:

1. Why not use memcached? (I can guess due to memcached expiration)
2. Why not use a database materialised view?
3. Why not rebuild the cached object on write? I see that now we just delete the cache entry and the entry will be rebuild at read time. This is how a materialised view will work.
4. The start-django script is run on each restart and deploy. Are we rebuilding the whole cache on each restart? This would mean a "2 minutes" outage on each restart/deploy. Note that this is a known issue with migrations.
5. For the future, serialised objects in external caches and blue/green deployments must be carefully handled.

punchagan:
> Few questions to understand the problem+solution:
> 
>     1. Why not use memcached? (I can guess due to memcached expiration)

Yes, an experimental `memcached` solution is already deployed and it works fine for smaller data sets on partner sites, but for the Akvo RSR page itself, it doesn't work because of the long delay in populating the cache. 

>     2. Why not use a database materialised view?

I mostly arrived here from trying to build a cache to see how fast it will be, and trying to optimize a view. It might be more efficient to use a materialised view, but the data being generated is quite complicated with a bunch of joins and stuff. I don't know how easy to understand and maintain the SQL is going to be.  I'm not super familiar with materialized views. Do you think there would be a significant benefit of using them? 

>     3. Why not rebuild the cached object on write? I see that now we just delete the cache entry and the entry will be rebuild at read time. This is how a materialised view will work.

There might be a bunch of writes which don't really change the actual cached data. We don't want to rebuild the cached object every time. 

>     4. The start-django script is run on each restart and deploy. Are we rebuilding the whole cache on each restart? This would mean a "2 minutes" outage on each restart/deploy. Note that this is a known issue with migrations.

No, we are not actually rebuilding the whole cache. I'm only running the command to create the DB table, if it is not already present.  We will need to run the cache refilling manually, if we change the format of the data being returned. There isn't a good way to do that, right now, given that the cache filling can take a while. 

>     5. For the future, serialised objects in external caches and blue/green deployments must be carefully handled.

Ah, good point. But, also it wouldn't be any different from working with any DB field, the way we are using this cache right now.

dlebrero:
Chatted with @punchagan offline.

We agreed:
1. Approach is good enough.
1. We will explore the materialised view option later, as it is more maintainable, makes reads more consistent (as there is not a random number of cache items to create) and removes some possible concurrency issues.
1. Table for cache creation should be part of DB migrations but we could not figure out how to do it. Not worth investing more time on this.

Issue #4205, opened 2020-05-28
New period endpoints


martinchristov:
### GET `/period-updates/:periodId`

returns a list of updates

### POST `/set-periods-locked`
payload: 
```
{ periods: [list of ids], locked: [bool] }

martinchristov:
Please use issue/3998-results branch as a basis

PR #4207, opened 2020-06-01
[#4205] new period endpoints

zuhdil:
- [ ] Documentation

### DIFF
--- a/akvo/rest/views/indicator_period.py
+++ b/akvo/rest/views/indicator_period.py
@@ -4,11 +4,16 @@
 # See more details in the license.txt file located at the root folder of the Akvo RSR module.
 # For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
 
-
-from akvo.rsr.models import IndicatorPeriod
-
 from ..serializers import IndicatorPeriodSerializer, IndicatorPeriodFrameworkSerializer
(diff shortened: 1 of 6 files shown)


akvo/rest/views/indicator_period.py
```diff
@@ -27,3 +32,23 @@ class IndicatorPeriodFrameworkViewSet(PublicProjectViewSet):
...
+    period_ids = request.data.get('periods', [])
+    locked = request.data.get('locked', None)
+    if len(period_ids) < 1 or locked is None:
+        return HttpResponseBadRequest()
+
+    user = request.user
+    periods = IndicatorPeriod.objects.filter(id__in=period_ids)
+
+    # only do the operation on periods where user has correct permission.
+    permitted_ids = [p.id for p in periods if user.has_perm('rsr.change_indicatorperiod', p)]
```
punchagan, line 50: This can be potentially very slow, with large projects. 

Currently, it is good enough to support indicators under a single project, and I think we should just enforce that with the API and check if the user has permissions to change indicator periods on that project. 

The permissions code itself can use some improvements to avoid doing these kind of checks hitting the DB, if we can, kind of like the recent optimisations I did for editable flags on projects, but we should think of a way to do this cleanly without adding a hackish optimisation on all such views.

PR #4213, opened 2020-06-02
Issue/4211 flags refactor

martinchristov:
Thought we need a more consolidated handling of feature flags

### DIFF
--- a/akvo/rsr/spa/app/utils/feat-flags.js
+++ b/akvo/rsr/spa/app/utils/feat-flags.js
@@ -1,0 +1,9 @@
+export const flagOrgs = {
+  FAC: new Set([42, 3210]),
+  CREATE_NEW_PROGRAM: new Set([42, 3394]),
+  CREATE_HIERARCHY_PROJECT: new Set([42, 3394])
+}
+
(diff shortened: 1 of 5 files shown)


akvo/rsr/spa/app/utils/feat-flags.js
```diff
@@ -0,0 +1,9 @@
+export const flagOrgs = {
+  FAC: new Set([42, 3210]),
+  CREATE_NEW_PROGRAM: new Set([42, 3394]),
+  CREATE_HIERARCHY_PROJECT: new Set([42, 3394])
+}
+
+export const shouldShowFlag = (orgs, orgSet) => {
+  return orgs && orgs.findIndex(it => orgSet.has(it.id)) !== -1
```
punchagan, line 8: I think we can include the `contentOwner` check here, directly. 

It seems like it makes sense for all the feature flags we have currently.  Specifically, I think we should show the Programs user menu even to users in the content owned organisations, which wouldn't happen with the way we have the flags configured, right now.

martinchristov: Cool. Actually I didn't really understand what is this property `contentOwner` before :-D

PR #4229, opened 2020-06-17
Fetch changes from GitHub before running promote

punchagan:
This adds about 3-4 seconds to the promote, but removes the need to run
the promote script from failing to send the list of commits as changes,
if the latest changes were not fetched locally already.

- [ ] Documentation


punchagan:
Yeah, that may be a better place to have this.  I'll update that script.

PR #4341, opened 2020-10-14
Added Cloud SQL proxy containers

dlebrero:
The containers will not be used but we need them in production for when we change the RSR backend DB configuration

See #4080


dlebrero:
my hub cli is not playing ball.

PR #4350, opened 2020-10-21
Make the project.update_iati_checks method faster for large projects

punchagan:
This PR has two commits, the first one makes perf improvements for projects with a large number of checks performed on them, essentially large projects with a lot of transaction/location/results/etc. 

The second commit reduces the number of DB queries made for projects with a large results framework. 

See commit messages for more details.


punchagan:
I added a 3rd commit, that is somewhat hackish optimisation, but should make the CRUD operations on individual objects much faster.

dlebrero (approved):
LGTM, but honestly I do not like the last commit but neither I liked the commit with the cache of the permissions.

zuhdil (approved):

PR #4363, opened 2020-10-28
Run IATI checks for projects asynchronously

punchagan:
- [ ] Documentation

This PR moves out the IATI check runs out of the project save to cron jobs. The checks are run every 5 minutes. We also run the IATI checks when data about the IATI checks for an organisation are fetched. This could potentially slow down the request for list of projects, but we hope that there aren't a lot of projects that have been updated in the last 5 minutes.


akvo/rest/views/project.py
```diff
@@ -151,6 +152,13 @@ def get_queryset(self):
             ).distinct()
         return super(ProjectIatiExportViewSet, self).get_queryset()
 
+    def list(self, request, *args, **kwargs):
+        projects = self.queryset.filter(run_iati_checks=True)
+        projects.update(run_iati_checks=False)
```
dlebrero, line 157: Do this after the update_iati_check or as part of it? Just in case there is some exception.

akvo/rsr/management/commands/perform_iati_checks.py
```diff
@@ -10,9 +10,21 @@
...
+            action='store_true',
+            default=False,
+            help='Run IATI checks for all the projects in the DB.',
+        )
 
     def handle(self, *args, **options):
-        for project in Project.objects.all():
+        projects = Project.objects.all() if options['all'] else Project.objects.filter(run_iati_checks=True)
+        if not options['all']:
+            projects.update(run_iati_checks=False)
```
dlebrero, line 26: See first comment.

dlebrero (changes requested):

akvo/rest/views/project.py
punchagan: Yes, moved it to be inside `update_iati_checks`. Thanks for the catch. I was trying to not miss out on checks that are "scheduled" while this loop is running. But, moving the change inside the loop should be good enough.

dlebrero (approved):
LGTM as long as we do not need to worry about concurrency on doing the checks (or it is a minor issue)

punchagan:
> LGTM as long as we do not need to worry about concurrency on doing the checks (or it is a minor issue)

Pushed a commit to perform the deletion of checks, creation of new checks, and marking project as checked in a transaction. This should make concurrency much less of an issue.

dlebrero:
Thanks.

FYI, I do not get an email from Github unless you click the "re-request review" button.

punchagan:
> FYI, I do not get an email from Github unless you click the "re-request review" button.

Thanks for letting me know. I did that for my other PR. :)

PR #4367, opened 2020-10-28
Initialize Sentry SDK with correct release name and environment in SPA

punchagan:
- [ ] Documentation

### DIFF
--- a/akvo/rsr/spa/app/root.jsx
+++ b/akvo/rsr/spa/app/root.jsx
@@ -20,7 +20,8 @@ import Profile from './modules/profile/profile'
 if (!(env && env.LOCALDEV)) {
   Sentry.init({
     dsn: 'https://5ca590f001844d1493e459a38e3d75f3@sentry.io/218288',
-    release: 'rsr-live@test-release'
+    release: window.SENTRY_RELEASE,
+    environment: window.SENTRY_ENVIRONMENT,
   })
 }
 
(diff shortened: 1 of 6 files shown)


akvo/rsr/spa/app/root.jsx
```diff
@@ -18,9 +18,11 @@ import IATI from './modules/iati/iati'
 import Profile from './modules/profile/profile'
 
 if (!(env && env.LOCALDEV)) {
+  const hostname = window.location.hostname
   Sentry.init({
     dsn: 'https://5ca590f001844d1493e459a38e3d75f3@sentry.io/218288',
-    release: 'rsr-live@test-release'
+    release: `rsr@${env.RELEASE}`,
+    environment: (hostname === 'rsr.akvo.org' || hostname.endsWith('akvoapp.org')) ? 'prod' : 'test'
```
dlebrero, line 25: This logic brings back bad memories from my old me.

Both the environment and the SHA are available (or could) as part of the Nginx env in k8s). Is there a reasonable option to use those?

We are already doing some "mangling" of files at the start up of the nginx process at https://github.com/akvo/akvo-rsr/blob/master/nginx/start.sh. Not that I like this mangling at all but maybe you can think of a better approach.

Otherwise, LGTM.

dlebrero (approved):

punchagan:
I've force pushed a couple of commits to set the sentry config using the file mangling in `nginx/start.sh`.

dlebrero (approved):
Again, I do not like the mangling approach either. 

Added some test.

PR #4368, opened 2020-10-29
Remove unnecessary build/index.html in spa and dir

punchagan:
The additional index.html file is confusing and doesn't seem to be really necessary with some changes to the webpack configuration. 

- [ ] Documentation


akvo/rsr/spa/webpack.config.js
```diff
@@ -26,7 +27,7 @@ const config = () => {
     output: {
       filename: 'bundle.js',
       path: resolve(__dirname, 'dist'),
-      publicPath: '',
+      publicPath: '/my-rsr/',
```
martinchristov, line 30: Why have you changed this?

akvo/rsr/spa/webpack.config.js
```diff
@@ -138,6 +138,11 @@ const config = () => {
 
     plugins: [
       new webpack.NamedModulesPlugin(),
+      new HtmlWebpackPlugin({
+        template: `${__dirname}/app/index.html`,
+        filename: 'index.html',
+        inject: 'body',
+      }),
```
martinchristov, line 145: Why have you added this in the dev config?

akvo/rsr/spa/webpack.config.js
punchagan: Because otherwise the links don't reach the webpack server. Our webpack servers + nginx configuration requires us to have URLs prefixed with /my-rsr/ or /project-directory/ to hit the webpack server.

akvo/rsr/spa/webpack.config.js
punchagan: Because we want to use the index.html and inject the bundle config in the dev environment. Why should it be a problem to add it to dev config?

punchagan:
@martinchristov does this work for you? If yes, I think we can merge this?

PR #4369, opened 2020-11-02
[#4356] Handle updating the disaggregation target values.

### LINKED ISSUE #4356: Improved FE/BE handling of S5 disaggregation target setting
The backend to handle updating of the disaggregation target values in the following manner:

PATCH `/indicator_period/494478/`
payload:
```
{"disaggregation_targets":[{"value":"32"},{"value":"23"}]}
```

preexisting target values update to be handled on an index basis perhaps

zuhdil:
- [ ] Documentation
 
Closes #4356

### DIFF
--- a/akvo/rest/serializers/indicator.py
+++ b/akvo/rest/serializers/indicator.py
@@ -6,7 +6,7 @@
 
 from akvo.rest.serializers.indicator_period import (
     IndicatorPeriodFrameworkSerializer, IndicatorPeriodFrameworkLiteSerializer,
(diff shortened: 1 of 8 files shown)


martinchristov:
Martin Christov: It's working as expected, unless for the special case for 2scale when targets are at indicator level

POST: http://localhost/rest/v1/indicator_framework/133612/
Payload:
{"disaggregation_targets":[{"value":"30"},{"value":"40"}],"project_editor_change":true}

PR #4376, opened 2020-11-10
Update IATI codelist v203 and generate JSON codelists

punchagan:
- [ ] Documentation

The PR makes some minor improvements to the codelist generator script, and then improves the actual codelists themselves. The PR also generates the JSON codelists that are hard-coded in the Project Editor SPA. 

This PR also fixes an unrelated bug with submitting empty CRS channel codes. This bug is not related to the IATI changes being made here, but was discovered in the process


zuhdil (approved):
LGTM, but I think it would be nice if the generated FE json lives on its own path/module as some files are reused by multiple section.

punchagan:
> LGTM, but I think it would be nice if the generated FE json lives on its own path/module as some files are reused in multiple sections.

That could come in a separate change. I've got rid of some of the duplicated files, and tried to keep the changes to a minimum so that it is easy to review this PR. We could if do the moving around of files, if we have the time and energy for it. Feel free to make an issue for it. :)

PR #4398, opened 2020-12-09
[#4397] Keep hierarchy org in sync with root project reporting org

### LINKED ISSUE #4397: Keep project hierarchy org and root project reporting org in sync
If not, we can have permission issues with users being able to see some programs but not being able to access them, etc.

punchagan:
Closes #4397

- [ ] Documentation


akvo/rsr/models/project_hierarchy.py
```diff
@@ -32,3 +34,21 @@ def project_count(self):
     @property
     def project_ids(self):
         return self.descendants.values_list('id', flat=True)
+
+
+@receiver(post_save, sender=ProjectHierarchy)
```
zuhdil, line 39: Hi @punchagan , is there any particular reason why to use the post_save? I mean, the organisation is being overridden anyway, which means the org param on create method will be ignored. So why not use the pre_save to that the org param no longer needed.

punchagan: Yes, `pre_save` makes sense. I was even considering getting rid of the field itself, on ProjectHierarchy. What do you think about that? In some of the edge cases, it might be useful to have an organisation on the ProjectHierarchy, but I wonder if we should just rely on the root project's reporting organisation and make this field a property.

zuhdil: I agree with that idea, by removing the field the `post_save` in the `Project` model is no longer needed, thereby more simple and less moving parts.

zuhdil (approved):

PR #4429, opened 2021-01-28
Add cors to include akvo.org subdomains

tangrammer:
Relates https://github.com/akvo/hortinvest/issues/5

- [ ] Documentation

Using https://nginx.org/en/docs/http/ngx_http_map_module.html

### DIFF
--- a/nginx/default.conf
+++ b/nginx/default.conf
@@ -8,11 +8,7 @@ server {
 }
 
 map $http_origin $allow_origin {
-    ~^https://(.*)+\.(akvo|akvotest)\.org$ $http_origin;
-}
-
-map $http_origin $allow_methods {
-    ~^https://(.*)+\.(akvo|akvotest)\.org$ "OPTIONS, HEAD, PUT, POST, GET, DELETE";
+    ~^https://.+\.(akvo|akvotest)\.org$ $http_origin;
 }
 
 server {
@@ -62,7 +58,7 @@ server {
 
         add_header   Front-End-Https   on;
         add_header Access-Control-Allow-Origin $allow_origin;
-        add_header Access-Control-Allow-Methods $allow_methods;
+        add_header Access-Control-Allow-Methods "*";
     }
 
     location /healthz {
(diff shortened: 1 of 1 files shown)


tangrammer:
realising that we should allow all methods ('PUT, POST, GET, DELETE, OPTIONS') only if rsr.akvo.org

punchagan:
We also have something called partner sites, which are basically subdomains of akvoapp.org and restrict the data being shown to a subset of RSR's data. For instance, https://eutf-syria.akvoapp.org/ https://unep.akvoapp.org/

tangrammer:
@punchagan but do these clients use AJAX to communicate with RSR API?

punchagan:
Yes, you can use the project editor etc.

dlebrero (changes requested):
What about the test environment?

dlebrero:
> @punchagan but do these clients use AJAX to communicate with RSR API?

But they do it through the ".akvoapp.org" domain, so they do not need cors headers.

nginx/default.conf
```diff
@@ -7,6 +7,14 @@ server {
   }
 }
 
+map $http_origin $allow_origin {
+    ~^https://(.*)+\.(akvo|akvotest)\.org$ $http_origin;
```
dlebrero, line 11: `(.*)+` can be simplified to .* or .+

nginx/default.conf
```diff
@@ -7,6 +7,14 @@ server {
...
+}
+
+map $http_origin $allow_methods {
+    ~^https://(.*)+\.(akvo|akvotest)\.org$ "OPTIONS, HEAD, PUT, POST, GET, DELETE";
```
dlebrero, line 15: I see no harm on adding it to all domains, so that we avoid the regex duplication

nginx/default.conf
```diff
@@ -53,6 +57,8 @@ server {
         limit_req zone=bots burst=2 nodelay;
 
         add_header   Front-End-Https   on;
+        add_header Access-Control-Allow-Origin $allow_origin;
+        add_header Access-Control-Allow-Methods "*";
```
dlebrero, line 61: Great!

dlebrero (approved):

PR #4480, opened 2021-04-07
Issue/4459 child period dates changeable

zuhdil:
- [ ] Documentation


akvo/rsr/migrations/0205_labeled_period_report.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
```
punchagan, line 1: This would need to be re-numbered because we already have another migration on master

punchagan (approved):
I'm a little skeptical of the performance. It takes about a minute to generate the report, even without any labels for OKP. But, I guess the performance would be similar even if we had labels, right? I guess this may be acceptable for now, and we could look at performance once there is more data?

PR #4496, opened 2021-04-22
[#4495] Project updates with multiple photos

zuhdil:
- [ ] Documentation


ifirmawan:
**Summary**
I made change to add new field ** Photos** which able to upload multiple photos  at Form Updates when we go to details Project.

**How it's work?**

1. Login to your Akvo RSR's account.
2. Select a project that will be seen in detail.
3. Go to **Updates** tab 
4. Fill in the title along with other fields on the form on the right.
5. In the Photos section click the ```+ Add photo``` button.
6. Click on the plus icon box.
7. Fill in the caption and credit on each photo then.
8. Click the ```Add an update``` button to send data to the server.

punchagan (changes requested):
Thanks for your work on this @ifirmawan and @zuhdil. 

I did some testing and found a few issues with the PR.  Here are my notes on it. Please let me know if you'd like more information, or if you'd like help with fixing them. 

1. Images uploaded without caption or credits show "undefined" as caption, when trying to edit them after the update has been published. 
2. Editing a published Update and trying to add new images doesn't seem to work correctly. The newly added image is ignored. 
3. Deletion of images in an update seems to be happening immediately on clicking the trash icon, and not on clicking Save update.  In this case we should show the user a pop-up confirmation dialog, and not directly delete the images. 
4. The published view of the update shows only the first (?) picture in the list. We should show all the photos uploaded.

ifirmawan:
Thank you for your review @punchagan . 
I just realized the published view of the update (issue number 4) doesn't include in SPA. but it's a part of Django views. I will investigate the end point of details updates with @zuhdil.

ifirmawan:
I try to resolve some feedbacks here, with commit #c52acd6071fd69626c5daa084727fd09e44e9c64
- [  ] 1. Images uploaded without caption or credits show "undefined" as caption, when trying to edit them after the update has been published. 
- [ ] 2. Editing a published Update and trying to add new images doesn't seem to work correctly. The newly added image is ignored.

ifirmawan:
Ready for review with the last commit e3c087a

- [ ] 3. Deletion of images in an update seems to be happening immediately on clicking the trash icon, and not on clicking Save update.  In this case we should show the user a pop-up confirmation dialog, and not directly delete the images. 
- [ ] 4. The published view of the update shows only the first (?) picture in the list. We should show all the photos uploaded.

Issue #4437, opened 2021-03-02
Login page


martinchristov:
The objective is to create a login/register page within the react environment which is identical in functionality to what we currently have.
We want to do that because:

- the old login view is part of the old codebase and it is the last remaining view of that non-SPA environment
- many people are frustrated with the Nuffic requirement to lock users out of RSR after 3 failed login attempts. In the new login view we can change that for non-Nuffic users. So the current restriction would be enabled only for Nuffic's subdomain, whereas the rest of the users would either have no restriction, or perhaps a CAPTCHA after 3 failed login attempts.

The new view would also need to implement the Register view and the forgot password functionality. It would need 3 endpoints respectively and to set the cookie that's currently being set on the server-side.

PR #4505, opened 2021-04-29
[#4437] Resolved issue login page

ifirmawan:
### First step refactory
Create new login design
* Recreate routing private and public in root.jsx.
* Recreate reducers to handle login states.
* Add new modules auth and dashboard.
* Each components still in dummy or static mode.

- [ ] Documentation


ifirmawan:
## Issues Checklist

- [ ] * The password requirements on new registrations are more stringent than just a 12 char length. The backend sends a 400 response, but the FE doesn't display this information.

- [ ] * The form text marking mandatory fields is blinking while I type another text into another field. This is visually jarring.

- [ ] * Is the copy of the text on the left hand side something that we have elsewhere in RSR, or did you come up with it? I think people might have opinions on it. 😄 

- [ ] * Could you check the register and login pages with Firefox? It could be a problem with my bleeding edge FF, but forgot password doesn't seem to work because CSRF token doesn't get sent? Also, there are some visual/CSS glitches.

Issue #4507, opened 2021-04-30
Enumerator form indicator title & description


ifirmawan:
The Enumerator form don't see the 'Indicator title', which is important. And the heading 'Description' is not included as it is in the app.

PR #4508, opened 2021-04-30
[#4507] Bug fixing initial indicator description

ifirmawan:
Description doesn't match with the first title.
* create initial states for indicators.
* change condition of preview url params.
* setSelected indicator after filtering periods isPeriodNeedsReporting.

- [ ] Documentation

### DIFF
--- a/akvo/rsr/spa/app/modules/enumerators/enumerators.jsx
+++ b/akvo/rsr/spa/app/modules/enumerators/enumerators.jsx
@@ -1,11 +1,10 @@
 /* global window */
-import React, { useEffect, useState, useRef } from 'react'
+import React, { useEffect, useState } from 'react'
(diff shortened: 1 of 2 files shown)


punchagan:
@ifirmawan if you are working on a branch, it is nice to squash commits into single logical ones, wherever possible. I think the 3 commits here can be squashed into a single one, so that the commit history makes more sense to anyone looking at it in future.

ifirmawan:
noted @punchagan , I will do that in the future

punchagan:
You can do that with the current PR too. Rebase against the `master` and squash the commits into a single one.

ifirmawan:
I see, I'm trying @punchagan.

zuhdil (approved):

PR #4515, opened 2021-05-05
Resolved issue EUTF Syria reported

ifirmawan:
## First issue

Something odd is happening when trying to report an indicator value (Loading indicator at the save button).

## Second issue

Trying to set the target for a disaggregation label in section 5 of the Editor. There is an helpful _"Something went wrong message"_
- [ ] Documentation

### DIFF
--- a/akvo/rsr/spa/app/modules/editor/actions.js
+++ b/akvo/rsr/spa/app/modules/editor/actions.js
@@ -37,11 +37,8 @@ export const addSetItem = (sectionIndex, setName, item) => (dispatch, getState) 
     })
 }
 export const editSetItem = (sectionIndex, setName, itemIndex, itemId, fields) => (dispatch) => {
-  const inputFields = fields?.disaggregationTargets?.length > 0
-    ? { ...fields, disaggregationTargets: fields.disaggregationTargets.filter(item => (item?.value)) }
-    : fields
   dispatch({ type: actionTypes.EDIT_SET_ITEM, sectionIndex, setName, itemIndex, fields })
-  api.patch(`${getEndpoint(sectionIndex, setName)}${itemId}/`, inputFields, getTransform(sectionIndex, setName, 'request'), null, true)
+  api.patch(`${getEndpoint(sectionIndex, setName)}${itemId}/`, fields, getTransform(sectionIndex, setName, 'request'), null, true)
     .then(() => { dispatch({ type: actionTypes.BACKEND_SYNC }) })
     .catch((error) => { dispatch({ type: actionTypes.BACKEND_ERROR, error, sectionIndex, setName: `${setName}[${itemIndex}]`, response: error.response ? error.response.data : error, statusCode: error.response.status }) })
 }
--- a/akvo/rsr/spa/app/modules/editor/actions.js
+++ b/akvo/rsr/spa/app/modules/editor/actions.js
@@ -7,11 +7,11 @@ export const touchSection = sectionIndex => ({ type: actionTypes.TOUCH_SECTION, 
 export const checkValidation = (id, checked) => ({ type: actionTypes.CHECK_VALIDATION, id, checked })
 export const saveFields = (fields, sectionIndex, preventUpload) => (dispatch, getState) => {
   dispatch({ type: actionTypes.SAVE_FIELDS, fields, sectionIndex })
-  const {projectId} = getState().editorRdr
-  if(!preventUpload){
-  api.patch(`/project/${projectId}/`, fields, null, null, true)
-    .then(() => dispatch({ type: actionTypes.BACKEND_SYNC }))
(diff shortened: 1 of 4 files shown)


akvo/rsr/spa/app/modules/editor/actions.js
```diff
@@ -24,38 +24,40 @@ export const setStatus = (publishingStatus) => (dispatch, getState) => {
...
-    .then(({ data: {id, periods}}) => { dispatch({ type: actionTypes.ADDED_SET_ITEM, sectionIndex, setName, id, itemIndex, periods }) })
+    .then(({ data: { id, periods } }) => { dispatch({ type: actionTypes.ADDED_SET_ITEM, sectionIndex, setName, id, itemIndex, periods }) })
     .catch((error) => {
       dispatch({ type: actionTypes.BACKEND_ERROR, error, sectionIndex, setName: `${setName}[${setItems.length - 1}]`, response: error.response ? error.response.data : error, statusCode: error.response.status })
       dispatch({ type: actionTypes.REMOVED_SET_ITEM, failedAdd: true, sectionIndex, setName, itemIndex: setItems.length - 1, skipValidation: true })
     })
 }
 export const editSetItem = (sectionIndex, setName, itemIndex, itemId, fields) => (dispatch) => {
+  const inputFields = fields?.disaggregationTargets?.length > 0
+    ? { ...fields, disaggregationTargets: fields.disaggregationTargets.filter(item => (item?.value)) }
```
punchagan, line 41: We don't actually want to be filtering in the case when `item.value` is `null`. We want to retain the null values to unset values which were previously set. If a target was incorrectly set before, sending the `null` value unsets it. Filtering out those values would mean that we are not able to unset those values any more. 


Also, it looks like requests from different places seem to be sending different data. (before this filtering code is run). This is what the data looks like when I try to set a target on Project 7908 > **Result 1** > Indicator 1 > Period 1.  See below for another different example of data on the same project. 

```json
{
	"disaggregation_targets": [
		{
			"dimension_value": 973,
			"id": 26742,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 974,
			"id": 26743,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 975,
			"id": 26744,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 977,
			"id": 26745,
			"period": 383236
		},
		{
			"dimension_value": 976,
			"id": 26746,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 978,
			"id": 26747,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 979,
			"id": 26748,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 980,
			"id": 26749,
			"period": 383236,
			"value": 61000
		},
		{
			"dimension_value": 981,
			"id": 26750,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 982,
			"id": 26751,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 983,
			"id": 26752,
			"period": 383236,
			"value": null
		},
		{
			"dimension_value": 984,
			"id": 26753,
			"period": 383236,
			"value": null
		}
	]
}
```

This is what the data looks like when I try to set a target on Project 7908 > **Result 2** > Indicator 1 > Period 1. 

```json
{
	"disaggregation_targets": [
		{
			"dimension_value": 977,
			"value": "2"
		},
		null,
		null,
		{
			"dimension_value": 980,
			"value": "3"
		}
	]
}
```

punchagan (changes requested):
Thanks for working on this, @ifirmawan. But, I think we'd need to change the fix to accommodate deleting previously entered values. Please see my inline comment.

I merged the first commit you have for fixing the spinner for submit vs save draft.

Also, in general it would be easier to review if whitespace changes came as a separate commit, and are not mixed with the actual change itself. 

Also, I noticed that one of your commit didn't have the issue referenced correctly. I use a hook to automatically insert the reference: https://gist.github.com/punchagan/a9914dd62a42b2fea757312536e817ce
Feel free to use it. (Place it in `.git/hooks/prepare-commit-msg` and make sure it is executable)

punchagan:
I force pushed on this branch after merging your first commit. You should be able to `pull --rebase` onto your local branch to continue working with this branch.

punchagan:
Also, it would be nice for the commit messages to have a short summary line describing the issue that is fixed, and then a bigger explanation if required. Having things like "first issue", "second issue" in the commit would mean that anyone looking at the commit history would have to dig for the Issue/PR on GitHub to understand what the issue is. It is nice to keep the commits as self contained atomic units as possible.

akvo/rsr/spa/app/modules/editor/actions.js
punchagan, line 41: I think it would be enough to make sure all the entries in the `disaggregation_targets` array have a `value` and set them to `null` when the `value` field is missing. We of course filter out items which are null in the array.

PR #4543, opened 2021-05-26
[#4542] Apply recently added or removed default periods to existing indicators

zuhdil:
- [ ] Documentation


akvo/rsr/spa/app/modules/editor/section5/periods/defaults-modal.jsx
```diff
@@ -92,13 +92,23 @@ export const DefaultsModal = ({ visible, setVisible, periodFields, copyDefaults
                             />
                           </Item>
                         </Col>
+                        <Col span={2} style={{ paddingTop: '2em' }}>
```
punchagan, line 95: These changes seem entirely unrelated to the bug fix and the commit message. Generally it is a good practice to keep commits atomic. This becomes especially important when a commit attempts at fixing bugs that we don't entirely understand. We may want to revert or change the commit, etc.

punchagan (approved):

PR #4593, opened 2021-07-16
[#4592] change project parent

zuhdil:
- [ ] Documentation


akvo/rsr/usecases/change_project_parent.py
```diff
@@ -0,0 +1,144 @@
...
+    # change parents of RF items
+    change_candidates = get_rf_change_candidates(project, new_parent)
+    for key, candidates in change_candidates.items():
+        model, parent_attr, _ = RF_MODELS_CONFIG[key]
+        items = model.objects.filter(id__in=candidates.keys())
+        for item in items:
+            if verbosity > 1:
+                print(f"Change {key} parent of {item} to {candidates[item.id]}")
+            setattr(item, f"{parent_attr}_id", candidates[item.id])
+            item.save()
```
punchagan, line 122: Would it be better to use queryset `update` to instead of calling `item.save`? The models have complicated `save` methods that try to update children on save, etc. We could short-circuit that and avoid running all of that code, if we are manually taking care of the order, etc. 

Something like this piece of untested code below:

```python
    for key, candidates in change_candidates.items():
        model, parent_attr, _ = RF_MODELS_CONFIG[key]
        for item_id, target_id in candidates.items():
            if verbosity > 1:
                print(f"Change {key} parent of {item_id} to {target_id}")
            model.objects.filter(id__in=[item_id]).update(**{f"{parent_attr}_id": target_id})
```

punchagan (changes requested):
Thanks @zuhdil for working on this!  This functionality would be quite useful for our partners. And we could eventually think about making this available in the UI, after we have run it manually with careful observation a few times.

In the time I have available, I tried to review this and left a few comments. I have not carefully looked at the tests, but I presume they have been modeled on the previous tests to include all the previous edge cases, etc.

Thanks again, for taking out the time to work on this!

akvo/rsr/usecases/change_project_parent.py
```diff
@@ -0,0 +1,144 @@
...
+
+def get_leave(branch):
+    return get_leave(branch['children'][0]) if len(branch['children']) else branch['item']
+
+
+def make_target_parent_map(tree, project_attr, object_id, target_id):
+    target_parent_map = {}
+    for node in tree:
+        if len(node['children']) < 2:
+            continue
```
punchagan, line 67: Do we want to log this case, when running in verbose mode? It would point to potential data errors?

akvo/rsr/usecases/change_project_parent.py
```diff
@@ -0,0 +1,144 @@
...
+        children = node['children']
+        first, second = get_leave(children[0]), get_leave(children[1])
+        if first[project_attr] == object_id and second[project_attr] == target_id:
+            target_parent_map[first['id']] = second['id']
+        elif second[project_attr] == object_id and first[project_attr] == target_id:
+            target_parent_map[second['id']] = first['id']
+    return target_parent_map
```
punchagan, line 74: Do we want to log the else condition as a potential data error scenario?

akvo/rsr/usecases/change_project_parent.py
```diff
@@ -0,0 +1,144 @@
...
+                related_projects__related_project=project_id,
+                related_projects__relation=RelatedProject.PROJECT_RELATION_CHILD
+            ) | Q(
+                related_to_projects__project=project_id,
+                related_to_projects__relation=RelatedProject.PROJECT_RELATION_PARENT
+            )
+        ).values_list('id', flat=True).first()
+        if parent_id is None:
+            break
+        family.append(parent_id)
```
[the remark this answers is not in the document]
zuhdil: The checking for `None` is used to immediately stop the while loop, so the list should not have `None` as a member.

zuhdil:
Thank for the great review @punchagan !

akvo/rsr/usecases/change_project_parent.py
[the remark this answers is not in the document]
punchagan: Ah, right. Sorry. :)

akvo/rsr/usecases/change_project_parent.py
```diff
@@ -105,6 +109,12 @@ def get_rf_change_candidates(project, new_parent):
 
 
 def change_parent(project, new_parent, verbosity=0):
+    """
+    This function changes a project's parent including its Result Framework objects by traversing up the hierarchy
```
punchagan, line 113: Usually python docstrings have a short summary line followed by a longer description, similar to a git commit message. I'll update this docstring and update the commit.

punchagan (approved):

PR #4625, opened 2021-09-01
4624 eutf common output indicators

punchagan:
The PR adds a couple of commits:

- Add a simple descendants method for Results. The method is a "clone" of the projects method, but simpler.
- Add a program report for EUTF that filters the existing Results & Indicators table to include only data related to the "Common Output Indicators" result. We allow for some flexibility in changing the result title or choosing the result to include in the report by making the result ID a parameter in the URL. 

See the commit messages for a more detailed description.

### DIFF
--- a/akvo/rsr/models/result/result.py
+++ b/akvo/rsr/models/result/result.py
@@ -160,6 +160,18 @@ class Result(models.Model):
             projects[result.project.id] = result.project.title
         return projects
 
+    def descendants(self, depth=None):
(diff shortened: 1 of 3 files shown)


punchagan:
@zuhdil merging this, but this is possibly something you can review once you are back!

punchagan:
Add the missing migration file in this commit: https://github.com/akvo/akvo-rsr/commit/e9a95c9b9

PR #4678, opened 2021-10-27
[#4669] Replace docker service network mode

### LINKED ISSUE #4669: Replace docker service network mode
Our current `docker-compose.yml` uses `network_mode: service:mainnetwork` which triggers bugs in `docker-compose` v1:

 - Can't `docker-compose run web some command` when `web` service is running docker/compose#4548 
 - Can't forward ports on `web` service (or other services that use the service network mode) e.g `services.web.ports: [ "1919:1919" ]` doesn't work

These especially hinders work with PyCharm that needs to forward ports for the remote debugger and when indexing the python dependencies while `web` is running

MichaelAkvo:
This PR gets rid of `network_mode: service:XXX` in the docker-compose YAMLs for the local env. Reasons are explained in the related issue.

In order to do so, the nginx config had to be modified to point to the specific hosts instead of `localhost`. I haven't grokked the CI setup, so the service network mode was simply moved to the CI docker-compose for it to stay as it is.

It is not clear where the service network mode was chosen due to how kubernetes works in production or because it looked that way without docker (everything runs on localhost) and that was just transferred. The PR shouldn't impact production configuration as only dev files were changed and the CI still works.

Closes #4669

### DIFF
--- a/ci/build-semaphoreci.sh
+++ b/ci/build-semaphoreci.sh
@@ -75,9 +75,9 @@ docker build --rm=false -t eu.gcr.io/${PROJECT_NAME}/rsr-nginx:${CI_COMMIT} -f D
 
 log Starting docker-compose for end to end tests
 touch "log_docker_compose_ci_prod"
-docker-compose -p rsrciprod -f docker-compose.yaml -f docker-compose.ci.prod.images.yaml up -d --build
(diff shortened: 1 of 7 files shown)


MichaelAkvo:
So, the service network mode was chosen because of a possible, attempted introduction of keycloak and other proxies in production. It was dropped in the end, but the goal to target `localhost` stayed. Undoing that in production can be done later.

This PR only touches the local dev environment and is ready for review.

zuhdil (approved):

Issue #4660, opened 2021-10-12
Split web service for local environment


MichaelAkvo:
Currently, the `web` docker-compose service uses [`start-django.sh`](https://github.com/akvo/akvo-rsr/blob/master/scripts/docker/dev/start-django.sh) which starts 3 different processes in the background. This introduces these problems:

 - restarting other processes when only one is needed
 - throwing away logs (the `web` service only outputs stuff before the background processes are started)
 - inability to debug the processes (can't attach remote debugger nor read logs)

PR #4685, opened 2021-11-02
[#4660] Do not run migration in reports container

zuhdil:
When running `docker-compose up` with empty postgres volume, either the
web or the reports container will be terminated due to failure to run
the migration.

- [ ] Documentation


MichaelAkvo (approved):
LGTM!

PR #4686, opened 2021-11-02
[#4679] HMAC activation workflow

### LINKED ISSUE #4679: Deprecation(django-registration): Move away from model workflow
django-registration deprecated the model workflow in https://github.com/ubernostrum/django-registration/commit/61054528068ae063dbcfa46a5c61360750f978cc and https://github.com/ubernostrum/django-registration/commit/1d7d0f01a24b916977016c1d66823a5e4a33f2a0 .

As the [documentation]((https://django-registration.readthedocs.io/en/3.2/upgrade.html#removal-of-model-based-workflow)) states, we should move to the two-step activation workflow. That is introduced in v3.0 with v3.0.1 being a fix release.

# TODO

 - Check if we still need a registration form
 - Upgrade django-registration to v3.0.1
 - Switch to two-step activation workflow
 - Test registration

zuhdil:
- [ ] Documentation

Closes #4679

### DIFF
--- a/akvo/rsr/tests/views/test_account.py
+++ b/akvo/rsr/tests/views/test_account.py
@@ -505,7 +505,7 @@ class JsonAuthRegisterTestCase(BaseTestCase, CsrfTokenRequestMixin):
             'password2': 'passwdpasswdA1$',
         }
 
-        with patch('registration.models.RegistrationProfile.send_activation_email') as patched_send:
+        with patch('akvo.rsr.models.user.send_mail') as patched_send:
             response = self.c.post('/auth/register/', json.dumps(data), content_type='application/json', HTTP_X_CSRFTOKEN=csrftoken)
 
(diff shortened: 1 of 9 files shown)


akvo/rsr/tests/views/test_account.py
```diff
@@ -505,7 +505,7 @@ def test_valid_json_data(self):
             'password2': 'passwdpasswdA1$',
         }
 
-        with patch('registration.models.RegistrationProfile.send_activation_email') as patched_send:
+        with patch('akvo.rsr.models.user.send_mail') as patched_send:
```
MichaelAkvo, line 508: I know it already existed like this, but maybe we should use the [testing email service](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#email-services) the django provides here.

MichaelAkvo (changes requested):

Issue #4654, opened 2021-10-12
Endpoint performance


MichaelAkvo:
Endpoint performance has been a long-standing issue with some endpoints taking multiple seconds to complete.

This epic will track the most problematic endpoints and try to find either targeted or general improvements.

MichaelAkvo:
![Profile stats.png](https://images.zenhubusercontent.com/615f039319c1f837a3cdece3/302b9675-db02-4210-9e95-8a16b74ddda7)

After profiling the tests, some of the heaviest methods (excluding migrations) are:

 - akvo.rest.viewsets._projects_filter_for_non_privileged_users
 - akvo.rsr.models.model_querysets.project.ProjectQuerySet.of_partners
 - akvo.rsr.models.user.User.my_projects

My initial assumption was that there were too many models, but it seems like our implementation of hierarchies and the permission system are the cause of the lethargy.

PR #4707, opened 2021-11-17
[#4654] Endpoint performance: Add django-extensions to generate graph models and other things

MichaelAkvo:
https://django-extensions.readthedocs.io

It's not activated by default and can be used to generate a UML graph of all models (among other things)

#4654: Endpoint performance


zuhdil (changes requested):
Please resolve the conflict

PR #4719, opened 2021-11-24
[#4718]  nginx container won't startup locally without reports container

### LINKED ISSUE #4718: `nginx` container won't startup locally without `reports` container
When running `docker-compose up -d nginx`, `nginx` will fail to start because it can't resolve the `reports` container. A workaround is to actually start the container, but often, it will just be idling.

MichaelAkvo:
Fix #4718: `nginx` container can startup locally without `reports` container

The same format is applied to all `location` blocks, now `nginx` can start even all upstreams don't exist.

### DIFF
--- a/scripts/docker/dev/nginx/default.conf
+++ b/scripts/docker/dev/nginx/default.conf
@@ -8,6 +8,9 @@ server {
     proxy_read_timeout          600;
     send_timeout                600;
 
+    # Docker's internal resolver
+    resolver 127.0.0.11;
+
     location /my-rsr/ {
         proxy_pass   http://web-spa:8080/;
         proxy_set_header   Host             $host;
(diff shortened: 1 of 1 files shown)


scripts/docker/dev/nginx/default.conf
```diff
@@ -8,6 +8,9 @@ server {
     proxy_read_timeout          600;
     send_timeout                600;
 
+    # Docker's internal resolver
+    resolver 127.0.0.11;
```
zuhdil, line 12: Will it be consistent across all docker installations for all its versions?

zuhdil:
why not just modify the `docker-compose.override.yaml` by adding the reports *(and web)* containers in the `depends_on` attribute instead of modifying the Nginx configuration?

```yaml
nginx:
    depends_on:
      - web
      - reports
      - ...
```

I think it will be easier to understand and in line with what we want to solve

zuhdil (changes requested):

scripts/docker/dev/nginx/default.conf
MichaelAkvo: It looks like it's been there for at least 5 years https://github.com/moby/moby/blame/a77317882d010b884a9101c6ad0b2d7db141082f/libnetwork/sandbox.go#L136

MichaelAkvo:
I didn't want to add it to the `depends_on` because the reports container seems optional to me and doesn't have to be started all the time. Most of the time it will idle and do nothing since we don't use containers much.

Issue #4725, opened 2021-12-08
/rest/v1/project_directory performance


MichaelAkvo:
The endpoint is **unbelievably** slow and incredibly large.

```bash
$ time curl 'http://localhost/rest/v1/project-directory' > dev/null
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0 11.6M    0 52749    0     0    113      0 29:56:54  0:07:46 29:49:08 13131

real    7m46.442s
user    0m0.021s
```

Basically 9k projects are selected in production and for each of them a bunch more requests are made.

MichaelAkvo:
The problem in production seems to have come from `get_thumbnail` never ending and blocking the entire process.

It might still be possible to further improve performance by using `cache_page` on the endpoint and running the `populate_project_directory_cache`  django command in a cron.

A rework of the landing page is coming, so this might not be relevant anymore.

PR #4726, opened 2021-12-09
[#4725] quickfix: cache entire project directory request

MichaelAkvo:
The projects don't change often, so we should be able to cache them for a while.
Additionally, every hour we will regenerate their cached values (take ~7m right now).

So the quick fix is two-fold:

 - cache the entire result and keep for an hour
 - regenerate per project cache every hour

Once the page cache expires, queries should take ~10-20 seconds, but only once until the next hour.

Related to #4725: /rest/v1/project_directory performance

### DIFF
--- a/akvo/rest/cache.py
+++ b/akvo/rest/cache.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python3
+from typing import Dict, List
 
 from akvo.cache import cache_with_key, delete_cache_data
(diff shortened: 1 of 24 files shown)


MichaelAkvo:
It takes ~7m to cache all projects. It then takes ~10-20 seconds for the first request and < 1 second for cached requests.

About half the time is spent rendering the response and on a slow network more time is spent transferring the response.

PR #4750, opened 2021-12-22
[#4741] Fix Nuffic collaborate draft updates bug

### LINKED ISSUE #4741: Collaborate draft update bug

zuhdil:
- [ ] Documentation

Fixes #4741

### DIFF
--- a/akvo/rest/views/indicator_period_data.py
+++ b/akvo/rest/views/indicator_period_data.py
@@ -60,15 +60,22 @@ class IndicatorPeriodDataFrameworkViewSet(PublicProjectViewSet):
     serializer_class = IndicatorPeriodDataFrameworkSerializer
     project_relation = 'period__indicator__result__project__'
 
-    def get_queryset(self):
-        queryset = getattr(self, '_c_queryset', None)
-        if queryset is None:
-            queryset = super(IndicatorPeriodDataFrameworkViewSet, self).get_queryset()
-            queryset = IndicatorPeriodData.get_user_viewable_updates(
-                queryset, self.request.user
-            )
-            self._c_queryset = queryset
+    def get_object(self):
+        obj = get_object_or_404(self.get_queryset(), pk=self.kwargs['pk'])
+        # check whether the user has permission
+        viewables = IndicatorPeriodData.get_user_viewable_updates(
+            self.get_queryset().filter(pk=self.kwargs['pk']),
+            self.request.user
+        )
+        if viewables.count() == 0:
+            self.permission_denied(self.request)
+        return obj
 
+    def filter_queryset(self, queryset):
+        queryset = super().filter_queryset(queryset)
+        queryset = IndicatorPeriodData.get_user_viewable_updates(
+            queryset, self.request.user
+        )
         return queryset
 
     def perform_create(self, serializer):
--- a/akvo/rsr/tests/rest/test_indicator_period_data.py
+++ b/akvo/rsr/tests/rest/test_indicator_period_data.py
@@ -13,6 +13,7 @@ from datetime import date
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.contrib.admin.models import LogEntry, ADDITION, CHANGE, DELETION
 from django.contrib.contenttypes.models import ContentType
+from django.test import override_settings
 
 from akvo.rsr.models import (
     Partnership, Result, Indicator, IndicatorPeriod,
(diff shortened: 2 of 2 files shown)


akvo/rest/views/indicator_period_data.py
```diff
@@ -60,15 +60,22 @@ class IndicatorPeriodDataFrameworkViewSet(PublicProjectViewSet):
...
+    def get_object(self):
+        obj = get_object_or_404(self.get_queryset(), pk=self.kwargs['pk'])
+        # check whether the user has permission
+        viewables = IndicatorPeriodData.get_user_viewable_updates(
+            self.get_queryset().filter(pk=self.kwargs['pk']),
+            self.request.user
+        )
+        if viewables.count() == 0:
+            self.permission_denied(self.request)
+        return obj
```
MichaelAkvo, line 72: Is this necessary? Shouldn't `filter_queryset(...)` be enough? It would return a 404. Or is a 403 explicitly required?

akvo/rsr/tests/rest/test_indicator_period_data.py
```diff
@@ -1669,3 +1670,280 @@ def test_query_audit_trail(self):
...
+                'title': 'Result  #1',
+                'indicators': [{
+                    'title': 'Indicator #1',
+                    'periods': [{
+                        'period_start': date(2010, 1, 1),
+                        'period_end': date(2010, 12, 31),
+                    }]
+                }]
+            }]).build()
+        period = project.get_period(period_start=date(2010, 1, 1))
```
MichaelAkvo, line 1750: It looks like these belong in `setUp()`. They seem to be in every test.

akvo/rsr/tests/rest/test_indicator_period_data.py
```diff
@@ -1669,3 +1670,280 @@ def test_query_audit_trail(self):
             }
         ]
         self.assertEqual(expected, detail_response.data['audit_trail'])
+
+
+class IndicatorPeriodDataCollaborateDraftTestCase(BaseTestCase):
```
MichaelAkvo, line 1675: The tests need better comments. Either as a docstring for the test e.g

```py
def test_get_results_no_collab(self):
    """"
    When a user doesn't collaborate with another on results, then...
    """
```

or as comments in the code. `# Given`, `# When`, and `# Then` don't help understand the test as they are stating the obvious. I think they should sum up the non-obvious e.g

```py
        self.c.login(username=user1.email, password='password')
        # Explain why the setting is being overriden and maybe how it should affect the result
        with override_settings(NUFFIC_ROOT_PROJECT=project.object.id):
            response = self.c.get(f'/rest/v1/project/{project.object.id}/results_framework/?format=json')
```

I'll add other comments where things aren't clear and and a comment would help

akvo/rsr/tests/rest/test_indicator_period_data.py
```diff
@@ -1669,3 +1670,280 @@ def test_query_audit_trail(self):
...
+                }]
+            }]).build()
+        period = project.get_period(period_start=date(2010, 1, 1))
+        update1 = period.add_update(user2, value=1, status='A')
+        period.add_update(user2, value=2, status='D')
+        # When
+        self.c.login(username=user1.email, password='password')
+        response = self.c.get('/rest/v1/indicator_period_data_framework/?format=json', content_type='application/json')
+        # Then
+        updates = response.data['results']
```
MichaelAkvo, line 1757: Why is `['results']`? Other tests compare `updates = response.data['results'][0]['indicators'][0]['periods'][0]['updates']`

MichaelAkvo (changes requested):

akvo/rest/views/indicator_period_data.py
zuhdil: We have received several complaints about the inconsistency of the response code. Using the appropriate error code can be helpful in tracking future errors if any.

akvo/rsr/tests/rest/test_indicator_period_data.py
zuhdil: It's personal taste, but I don't really like using the setup method to make the test more "DRY". In a large test class, it often leaves me scrolling up and down just to understand the context of a test case.

I prefer the "DAMP" approach where I can see everything that matters to the test case in one place.

But I've made some changes in the test code to make it more "DRY"

akvo/rsr/tests/rest/test_indicator_period_data.py
zuhdil: I opted to write more descriptive method names.

I'm using the following format for the test method:
`{action}__{context_summary}__{expected_result}`

akvo/rsr/tests/rest/test_indicator_period_data.py
zuhdil: It is testing different endpoints. One is an endpoint to get the results framework of a project and the other is an endpoint to get the list of indicator period updates

akvo/rest/views/indicator_period_data.py
MichaelAkvo: I see :thinking:. There's now an issue where that can be handled #4759

akvo/rsr/tests/rest/test_indicator_period_data.py
MichaelAkvo: Hmm... that throws me back to Java with its long method names. We should probably agree upon something in order to keep it consistent, but that can be done later.

akvo/rsr/tests/rest/test_indicator_period_data.py
MichaelAkvo: OK. We should probably agree on something later to make it more consistent in future code.

akvo/rsr/tests/rest/test_indicator_period_data.py
```diff
@@ -1669,3 +1670,178 @@ def test_query_audit_trail(self):
...
+        return project, period
+
+    def enable_collab_feature_for(self, project):
+        return override_settings(NUFFIC_ROOT_PROJECT=project.object.id)
+
+    def test_get_results_framework__project_without_collab_feature__should_not_be_able_to_see_other_user_draft_update(self):
+        org = self.create_organisation('Acme')
+        user1, user2 = self.create_users(['user1', 'user2'], org)
+        project, period = self.make_project_with_one_period(org)
+        period.add_update(user2, value=1, status='D')
```
MichaelAkvo, line 1712: Could you use the constants e.g `IndicatorPeriodData.STATUS_DRAFT_CODE`  instead of literals e.g 'D'? It isn't immediately obvious what it means.

MichaelAkvo (changes requested):
Once the constants are used, it should be OK to merge.

PR #4787, opened 2022-01-12
[#4781]: Allow passing date range for perform_iati_checks

### LINKED ISSUE #4781: Allow passing date range for perform_iati_checks
Currently, [`perform_iati_checks`](https://github.com/akvo/akvo-rsr/blob/master/akvo/rsr/management/commands/perform_iati_checks.py) only has the `--all` option. It should be possible to pass a date range of projects to execute the script on.

Executing it on all projects is unnecessary as most are probably inactive.

This is related to #4779 as some project will have to be rechecked and we don't want to check all projects.

MichaelAkvo:
The only options we had were checking all projects or those still to be checked.
Now we can also check all projects before, after or in between dates.

Closes #4781

### DIFF
--- a/akvo/rsr/management/commands/perform_iati_checks.py
+++ b/akvo/rsr/management/commands/perform_iati_checks.py
@@ -3,25 +3,60 @@
 # Akvo Reporting is covered by the GNU Affero General Public License.
 # See more details in the license.txt file located at the root folder of the Akvo RSR module.
 # For additional details on the GNU license please see < http://www.gnu.org/licenses/agpl.html >.
+import argparse
+import datetime
 
 from akvo.rsr.models.project import Project
 
 from django.core.management.base import BaseCommand
 
 
+def date_arg_type(string):
+    return datetime.datetime.strptime(string, '%Y-%m-%d').date()
+
+
 class Command(BaseCommand):
     help = "Perform all IATI checks for projects."
 
-    def add_arguments(self, parser):
+    def add_arguments(self, parser: argparse.ArgumentParser):
         parser.add_argument(
             '--all',
             action='store_true',
             default=False,
             help='Run IATI checks for all the projects in the DB.',
+
+        )
+        parser.add_argument(
+            '--date-start',
+            type=date_arg_type,
+            help='Limit to projects created on and after this day',
+        )
+        parser.add_argument(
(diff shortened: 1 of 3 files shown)


akvo/rsr/management/commands/tests/test_perform_iati_checks.py
```diff
@@ -0,0 +1,79 @@
...
+        self.run_command("--date-start", "2010-10-6")
+        self.assertIn("Filtering projects after", self.stdout.getvalue())
+
+        # Half of the project should've been checked
+        self.assertEqual(Project.objects.filter(run_iati_checks=True).count(), 5)
+
+    def test_option_date_end(self):
+        """Projects after the given date should be checked"""
+
+        self.run_command("--date-start", "2010-10-6")
```
zuhdil, line 65: this should be `--date-end` right?

akvo/rsr/management/commands/perform_iati_checks.py
```diff
@@ -3,25 +3,60 @@
...
+            projects = projects.filter(run_iati_checks=True)
+        elif all_option:
+            self.stdout.write("Checking ALL projects. This might take a while...")
+        else:
+            if date_start:
+                self.stdout.write("Filtering projects after %s" % date_start)
+                projects = projects.filter(created_at__gte=date_start)
+            if date_end:
+                self.stdout.write("Filtering projects before %s" % date_end)
+                projects = projects.filter(created_at__lte=date_end)
```
zuhdil, line 58: "Filtering projects before `date_end`", so I assume projects created at date_end are not included, `created_at__lt=date_end` is more appropriate IMO

zuhdil (changes requested):

akvo/rsr/management/commands/perform_iati_checks.py
MichaelAkvo: Woops, the logs are incorrect. I do want it to include that date. I'll update the logs

akvo/rsr/management/commands/tests/test_perform_iati_checks.py
MichaelAkvo: :woozy_face: you're right!

Issue #4914, opened 2022-03-23
Bug: Description field text - Aqua for All


ifirmawan:
### What were you doing?

The issue is thus, that text was changed without anyone typing it and apparently copied from a new project she had open. Our comms intern thinks it’s a bug. I think so too as we checked all other options.

### What should've happened?

The description text should not change when there is no user update

PR #4915, opened 2022-03-24
[#4914] Fix bug description field at project editor

ifirmawan:
# TODO / Done

 - [x] Modify state fetching flag (boolean) for section 4 at reducer to make green check indicator at left side menu tied with the component from the right side.

# Test plan

Login as Project Editor eg: Janine de Laat (j.del***@aquaforall.org) and go to following projects
 - [ ] [WA-KE UP: Accelerating Water Access in Kenya](http://localhost/my-rsr/projects/9615/info) go to quickly to Editor -> Descriptions section and copy it to keep that in mind. Then go back and try to go to another project
 - [ ] [Aggregator for filter sales in Amhara Region, Ethiopia](http://localhost/my-rsr/projects/9284/info) go to quickly to Editor -> Description section and see what we have. is it the same as before? if the answer is yes then it's a bug.


MichaelAkvo (approved):
Looks good to me

PR #4956, opened 2022-05-10
[#4905] Implement project_published_search endpoint

### LINKED ISSUE #4905: Backend: Provide public project search endpoint
The new landing page will move away from searching a local cache to calling a search endpoint on the sever.

`/rest/v1/project_published_search`

# Specs

 - Only public and published projects should be searched
 - There should be a server-side limit on the number of projects that can be returned (django setting = 100)
 - Clients should be able to pass the max number of projects (limited by the server of course)

## Params

 - `query` - `String` **Optional**: the filter to use when finding projects - replicates what the frontend used to do
 - `sectors` - `List[String]` **Optional**: A list of sectors that the projects should be in (`OR`)
 - `orgs` - `List[Integer]` **Optional**: A list of organisation IDs that the the projects should be belong to (`OR`)
 - `limit` - `Integer` **Optional**: Max number of projects with data

## Result structure

```js
{
  "total": "<int>",
  // IDs of projects found
  "results": [
    "id": "<int>"
  ]
}
```

# TODO

 - Check if such an endpoint already exists or whether and existing endpoint can be modified to fit the specs
 - Add django setting `AKVO_PUBLIC_PROJECT_SEARCH_LIMIT` with an initial limit of 100
 - Add client side params

zuhdil:
# TODO / Done

Implement `project_published_search` endpoint

# Test plan

 - [x] Unit tests

Closes #4905


akvo/rest/views/project.py
```diff
@@ -331,6 +331,23 @@ def projects_by_id(request):
     return Response(serializer.data)
 
 
+@api_view(['GET'])
+def project_published_search(request):
+    query = request.GET.get('query', '')
+    sectors = {sector for sector in request.GET.get('sectors', '').split(',') if sector}
+    orgs = {int(org) for org in request.GET.get('orgs', '').split(',') if org}
+    limit = request.GET.get('limit', '')
+    limit = int(limit) if limit else settings.AKVO_PUBLIC_PROJECT_SEARCH_LIMIT
```
MichaelAkvo, line 340: The limit should be `1 <= limit <= AKVO_PUBLIC_PROJECT_SEARCH_LIMIT`

MichaelAkvo (changes requested):

ifirmawan (approved):

Issue #4939, opened 2022-04-25
[Hierarchies] Iati import


MichaelAkvo:
When importing projects from IATI, they can be set as the child of another one. This might be broken and has to be handled.

PR #4964, opened 2022-05-18
[#4939] Remove IATI import feature

MichaelAkvo:
# TODO / Done

Remove all references to the IATI import feature. It was decided not to continue maintaining the feature as it's not used at all.

# Test plan

Ensure that the unit tests pass

Related to #4939.  
First part of #4950


zuhdil (changes requested):
The cronjob for the `iati_import` task also needs to be removed 
https://github.com/akvo/akvo-rsr/blob/5df9ff75d2c129e444e249c400988249ca158c62/akvo/settings/90-finish.conf#L105

zuhdil:
LGTM!

zuhdil (approved):

PR #5018, opened 2022-06-28
[#5017] Speed up Results framework endpoint

### LINKED ISSUE #5017: Bug: Results framework takes a long time to load
### What were you doing?

Navigated to a project with a large results framework. It took ~10s to load.

E.g https://rsr.akvo.org/my-rsr/projects/10253/results takes 8,8s to load.

### What should've happened?

It should've loaded much faster

### My environment

# Desktop

OS: nixOS 21.11

Browser: Firefox 99.0.1 (64-bit)

### Additional context

After an initial investigation, the endpoint was not opitmised to prefetch related django models. ~1300 DB requests were being made that took about 2 seconds alone.

MichaelAkvo:
# TODO / Done

Prefetch as much data as possible.

`IndicatorPeriodFrameworkNoToLiteSerializer.get_updates` wasn't taking advantage of the prefetched data.
Now, it filters in python without making an extra query.
It's still suboptimal, but at least it's a bit faster.

# Test plan

 - [x] Manual testing

----------------

Closes #5017

### DIFF
--- a/akvo/rest/serializers/indicator.py
+++ b/akvo/rest/serializers/indicator.py
@@ -144,7 +144,7 @@ class IndicatorFrameworkNotSoLiteSerializer(BaseRSRSerializer):
     periods = IndicatorPeriodFrameworkNotSoLiteSerializer(many=True, required=False, read_only=True)
     parent_indicator = serializers.ReadOnlyField(source='parent_indicator_id')
     children_aggregate_percentage = serializers.ReadOnlyField()
-    labels = LabelListingField(read_only=True)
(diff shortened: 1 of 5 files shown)


zuhdil (changes requested):
I don't think the dimension_names attribute is unnecessary. Quick grep on the frontend code yield 15 matches
Screenshot_2022-06-29T11:26:28 168045850

zuhdil (commented):
@MichaelAkvo can you also update the commit message, I'm afraid it may cause confusion.

PR #5023, opened 2022-06-30
Speed up indicator_data_period_framework endpoint

MichaelAkvo:
# TODO / Done

The UserDetailsSerializer adds many queries to get the approved_organisations.
From what I can see, the caller of POST/PATCH `indicator_data_period_framework` doesn't even consume the user data.
It would be great to remove it entirely, but it's not clear if the field is used elsewhere
 and the UserRawSerializer remove still provides some information.

After testing the Results admin to POST and PATCH, the operations took 1.5s instead of 16s.

# Test plan

 - [x] Manual testing of Results Overview and Results admin for project [10355 ](http://localhost/my-rsr/projects/10355/results)
    - [x] Add new data using the results admin
    - [x] Reload page and delete data from Results Overview

Unlocking a period might be necessary
image


MichaelAkvo:
Not safe. Maybe for consideration later on. The data is changed, which might break the frontend. #5024 should have the same speed up effect

MichaelAkvo:
Will probably never look at this again, so time to close it.

PR #4970, opened 2022-05-25
Remove dependency on `related_projects`

MichaelAkvo:
# TODO / Done

Summarize what has been changed / what has to be done in order to finalize the PR.

 - [x] Remove `related_project` endpoint
 - [ ] Ensure the app works without `RelatedProjects`
 - [x] Add endpoint to add children to projects (for adding projects that "Contributes to") #5042
 - [x] Update `/v1/project` endpoint to accept setting parent contributing project #5042
 - [x] Set parent contributing project using `/v1/project` endpoint
 - [x] Return `Project.path` in `ProjectSerializer`

# Test plan

What tests are necessary to ensure this works or doesn't break anything working

 - [ ] Manual: Project editor should open
 

Closes #4969: [Hierarchies] FE Remove dependency on `related_projects`
Closes #4655: /rest/v1/related_project/ endpoint performance

PR #5042, opened 2022-07-18
[#5041] Migrate "Contributing Projects" to new project hierarchies

### LINKED ISSUE #5041: [Hiearchies] Migrate "Contributing Projects"
# Situation

## Child contributing projects

A project on RSR can have external projects that contribute to it. These were thus seen as external, child, contributing projects.

In the old data model, `RelatedProject` takes care of this relationship. Such children have

 - `related_iati_id` set to a non-empty string
 - `relation` set to `PROJECT_RELATION_CHILD`

## Parent contributing projects

Projects in RSR can also themselves contribute to a project. The target project is thus a parent contributing project. 

In the old data model, this relationship is stored in `RelatedProject`. These parents can be internal (`RelatedProject.related_project is not None` model) or external (`RelatedProject.related_project is None`). In both cases the other fields are the same:

 - `related_iati_id` set to a non-empty string
 - `relation` set to `PROJECT_RELATION_PARENT`

# TODO

Both these relationships have to be migrated away from `RelatedProject` as it is being phased out.

MichaelAkvo:
# TODO / Done

 - [x] Migrate child contributing projects `ExternalProject`
 - [x] Change frontend call to add "contributing external child project"
 - [x] Migrate internal parent contributors to `Project.contributes_to_project`
 - [x] Migrate external parent contributors to `Project.external_parent_iati_activity_id`
 - [x] **Move migrating project hierarchy out of `./manage.py migration`. It depends on methods that don't work with [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models)**

The frontend part for parent contributing projects is planned to be done in #4970 since the endpoint is being called there

# Test plan

 - [x] Unit tests
 - [x] Manual test in frontend to add contributing external child project
 

# Result

image

Closes #5041


MichaelAkvo:
@ifirmawan I don't know if you'd like to do the frontend part for parent contributing projects in this PR or https://github.com/akvo/akvo-rsr/pull/4970 .

Right now, all I have is the `/v1/project` endpoint which you can use to set the `contributes_to_project` to set the contributing parent if it's an RSR project, or `external_parent_iati_activity_id` if the project is not in RSR. The endpoint should be called with `PATCH` for a partial update and will return a 200 if it succeeds.

zuhdil (commented):
I ran into the following error while running the migration over a local production DB dump.

Screenshot_2022-07-26T09:50:50 035206807

MichaelAkvo:
Weird... I'm not getting that

MichaelAkvo:
Fixed.
Moved migrating project hierarchy out of `./manage.py migration`. It depends on methods that don't work with [historical models](https://docs.djangoproject.com/en/3.2/topics/migrations/#historical-models)

It now has to be fun manually with `./manage.py migrate_related_project_hierarchies`

zuhdil (approved):

ifirmawan:
> @ifirmawan I don't know if you'd like to do the frontend part for parent contributing projects in this PR or #4970 .
> 
> Right now, all I have is the `/v1/project` endpoint which you can use to set the `contributes_to_project` to set the contributing parent if it's an RSR project, or `external_parent_iati_activity_id` if the project is not in RSR. The endpoint should be called with `PATCH` for a partial update and will return a 200 if it succeeds.

Hi @MichaelAkvo  sorry just got around to reply. Let me try to do some work and let's see what we have after that.

Issue #5051, opened 2022-07-22
Feature Request: Full decline history when indicator update is returned for revision


AkvoAnt:
### What are you trying to do?

When an M&E manager declines a submitted indicator update, from the Pending Approval tab, the comment, i.e., the reason for the 'Return for revision', is currently not included / displayed as part of the 'Audit trail':

![image](https://user-images.githubusercontent.com/26227538/180453908-2ec34560-a162-4b8d-94fb-15f208de4618.png)

### Describe the solution you'd like

Include the reason for the 'Return for revision' text as part of the indicator 'Audit trail'.

PR #5056, opened 2022-08-01
[#5051] Show review note if status is R in Audit trail

ifirmawan:
# TODO / Done

 - [x] Show comment / review not if status is Revision in Audit trail

# Test plan

 - [ ] Check audit trails on this project : http://localhost/my-rsr/projects/9604/results

# Check audit trails
1. Login as MnE manager
2. Go to project http://localhost/my-rsr/projects/9604/results
3. Find indicator title : Number of networks meetings on sustainable WASH elements
4. Open first update which is value 222,222
5. Open Audit trail panel and find any declined status, the comment should be there.

### DIFF
--- a/akvo/rsr/spa/app/components/AuditTrail.jsx
+++ b/akvo/rsr/spa/app/components/AuditTrail.jsx
@@ -50,6 +50,12 @@ const DetailsValue = ({ data, textReport, disaggregations, scores }) => {
           </Col>
         </>
       ) : null}
+      {(data?.status === 'R' && data?.reviewNote) && (
+        <>
+          <Col span={span[0]}><Text type="secondary">Notes</Text></Col>
+          <Col span={span[1]}>{data.reviewNote}</Col>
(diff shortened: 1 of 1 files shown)


akvo/rsr/spa/app/components/AuditTrail.jsx
```diff
@@ -50,6 +50,12 @@ const DetailsValue = ({ data, textReport, disaggregations, scores }) => {
           </Col>
         </>
       ) : null}
+      {(data?.status === 'R' && data?.reviewNote) && (
+        <>
+          <Col><Text type="secondary">Comment</Text></Col>
+          <Col>{data.reviewNote}</Col>
```
zuhdil, line 56: I think the `Col` elements are missing the `span` attribute

ifirmawan: ah Okay, let me try to fix it

Issue #5062, opened 2022-08-03
Bug: Generating an IATI report for a large org uses a lot of memory


MichaelAkvo:
### What were you doing?

For a large organisation generate an [IATI export](https://rsr.akvo.org/my-rsr/iati) and wait.
Look at the memory usage as it increases by multiple GB (in my case 6GB for an org with ~630 projects)

### What should've happened?

The memory usage should either stay stable or only increase by a few MB, not GB. The generated XML is only 8MB.

MichaelAkvo:
This is currently "resolved" by increasing the memory of the K8s node

PR #5072, opened 2022-08-11
devops: place rsr on its own node

MichaelAkvo:
# TODO / Done

RSR has just become quite CPU intensive and requires more memory than before.
As a workaround before optimizing performance and memory, the pod will:
 - tolerate a mode with the taint akvo-app=rsr
 - have an affinity for nodes with the label akvo-app=rsr

# Test plan

Can't really test this without running it. Only a dry run can check the syntax.

Related to #5062

### DIFF
--- a/ci/k8s/deployment.yml
+++ b/ci/k8s/deployment.yml
@@ -22,6 +22,21 @@ spec:
       annotations:
         prometheus.io/scrape: 'true'
     spec:
+      tolerations:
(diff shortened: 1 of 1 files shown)


MichaelAkvo:
The syntax looks OK

```sh
$ kubectl apply --dry-run=client -f deployment.yml 
deployment.apps/rsr configured (dry run)
```

I'll merge this and see how it does on test.

Issue #5073, opened 2022-08-12
Bug: Editor not loading/Long delays before loading


ifirmawan:
### What were you doing?

On Firefox there are long delays but the editor eventually loads. This specifically happens when you navigate to the page directly e.g. [https://rsr.akvo.org/my-rsr/projects/10397/contacts](http://url/) Screenshot of what the users are experiencing

![image](https://user-images.githubusercontent.com/20871229/184268973-d253a436-91bf-4de8-aeaa-086761119fd7.png)

### What should've happened?

project editor must hide loading indicator after data is loaded.

ifirmawan:
![Uploading Screen Shot 2022-10-21 at 18.42.13.png…]()

ifirmawan:
**Invalid timezone in certain project**

<img width="613" alt="Screen Shot 2022-10-21 at 19 34 35" src="https://user-images.githubusercontent.com/20871229/197196799-8f040000-7218-4b1d-b1aa-6781140f9899.png">

MichaelAkvo:
Fixed long ago

PR #5074, opened 2022-08-12
[#5073] Fix editor long delays before loading

ifirmawan:
# TODO / Done

 - [x] Replace uncontrolled **nextSectionIndex** with controlled state in project editor.

# Test plan

 - [x] Open general information directly by following <a href="http://localhost/my-rsr/projects/2/info" target="_blank">this URL</a>
 - [x] Open project contacts directly by following <a href="http://my-rsr/projects/2/contacts" target="_blank">this URL</a>
 - [x] Open partner & user access directly by following <a href="http://localhost/my-rsr/projects/2/partners" target="_blank">this URL</a>
 - [x] Open descriptions directly by following <a href="http://localhost/my-rsr/projects/2/descriptions" target="_blank">this URL</a>
 - [x] Open results & indicators directly by following <a href="http://localhost/my-rsr/projects/2/results-n-indicators" target="_blank">this URL</a>
 - [x] Open finance directly by following <a href="http://localhost/my-rsr/projects/2/finance" target="_blank">this URL</a>
 - [x] Open location directly by following <a href="http://localhost/my-rsr/projects/2/locations" target="_blank">this URL</a>
 - [x] Open project focus directly by following <a href="http://localhost/my-rsr/projects/2/focus" target="_blank">this URL</a>
 - [x] Open project links & documents by following <a href="http://localhost/my-rsr/projects/2/links-n-docs" target="_blank">this URL</a>
 - [x] Open comments & keywords directly by following <a href="http://localhost/my-rsr/projects/2/comments-n-keywords" target="_blank">this URL</a>
 - [x] Open CRS++ & FSS reporting directly by following <a href="http://localhost/my-rsr/projects/2/reporting" target="_blank">this URL</a>

### DIFF
--- a/akvo/rsr/spa/app/modules/editor/project-init-handler.js
+++ b/akvo/rsr/spa/app/modules/editor/project-init-handler.js
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useState } from 'react'
 import { connect } from 'react-redux'
 import api from '../../utils/api'
 import { endpoints, getTransform } from './endpoints'
@@ -11,10 +11,15 @@ const insertRouteParams = (route, params) => {
   })
   return route
 }
-let prevParams
-let sectionIndexPipeline = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
+
+const SECTION_RESULTS_INDICATORS = 4;
+
+let sectionIndexPipeline = [1, 2, 3, SECTION_RESULTS_INDICATORS, 5, 6, 7, 8, 9, 10, 11];
+const SECTION_COUNT = sectionIndexPipeline.length + 1;
 
 const ProjectInitHandler = connect(({editorRdr}) => ({ editorRdr }), actions)(React.memo(({ match: {params}, ...props}) => {
+  const [nextSectionIndex, setNextSectionIndex] = useState(0)
+
   const fetchSection = (sectionIndex) => new Promise(async (resolve, reject) => {
     if(sectionIndex === 4 || sectionIndex === 6 || sectionIndex === 7 || sectionIndex === 8 || sectionIndex === 10){
       props.fetchSectionRoot(sectionIndex)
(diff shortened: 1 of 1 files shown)


akvo/rsr/spa/app/modules/editor/project-init-handler.js
```diff
@@ -64,26 +66,30 @@ const ProjectInitHandler = connect(({editorRdr}) => ({ editorRdr }), actions)(Re
...
+      if (!props?.editorRdr[`section${nextSectionIndex + 1}`]?.isFetched) {
+        fetchSection(sectionIndexPipeline[nextSectionIndex])
+          .then(() => {
+            if (nextSectionIndex === 4) {
+              /**
+               * Results and indicator need 10s delay to avoid fields component errors.
+               */
+              setTimeout(() => {
+                props.setSectionFetched(sectionIndexPipeline[nextSectionIndex])
+              }, 10000)
```
MichaelAkvo, line 82: Why are there field component errors? If it's too long for this PR, then a new issue should be created to treat this later and the issue number mentioned in this comment

akvo/rsr/spa/app/modules/editor/project-init-handler.js
```diff
@@ -64,26 +66,30 @@ const ProjectInitHandler = connect(({editorRdr}) => ({ editorRdr }), actions)(Re
...
-    })
-  }
+
   useEffect(() => {
-    if (prevParams && prevParams.id !== params.id && params.id !== 'new'){
-      fetchSection(3)
-      fetchSection(5)
+    // eslint-disable-next-line no-restricted-globals
+    if (!isNaN(params?.id) && (nextSectionIndex < 11)) {
+      if (!props?.editorRdr[`section${nextSectionIndex + 1}`]?.isFetched) {
```
MichaelAkvo, line 73: Looks like these can be put into one `if`.

--------------

Also, maybe it's my programming style, but I think something like this looks better

```js
if(badCondition){
  return
}
// initialize variables for next condition
if(otherBadCondition){
  return
}
// actual body of the function
```

Results in less indentation and clear rules with no nested `if`s for the reader to figure out when the function should be doing something.

-----------

Finally, it would be great to use a constant here instead of just `11` (e.g `SECTION_COUNT`). Probably it could be used elsewhere too (not for this PR).

akvo/rsr/spa/app/modules/editor/project-init-handler.js
```diff
@@ -64,26 +66,30 @@ const ProjectInitHandler = connect(({editorRdr}) => ({ editorRdr }), actions)(Re
...
+        fetchSection(sectionIndexPipeline[nextSectionIndex])
+          .then(() => {
+            if (nextSectionIndex === 4) {
```
MichaelAkvo, line 76: A constant here too instead of `4` would be better e.g `SECTION_RESULTS_INDICATORS`.

akvo/rsr/spa/app/modules/editor/project-init-handler.js
```diff
@@ -11,10 +11,12 @@ const insertRouteParams = (route, params) => {
   })
   return route
 }
-let prevParams
+
 let sectionIndexPipeline = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
 
 const ProjectInitHandler = connect(({editorRdr}) => ({ editorRdr }), actions)(React.memo(({ match: {params}, ...props}) => {
+  const [nextSectionIndex, setNextSectionIndex] = useState(0)
```
MichaelAkvo, line 18: I know nothing of react, so maybe this is OK, but state `0`? Will it always be `0` and how do we know it will stay `0`?

MichaelAkvo (changes requested):

akvo/rsr/spa/app/modules/editor/project-init-handler.js
ifirmawan: Thanks @MichaelAkvo , so Actually I'm trying to move from uncontrolled state to controlled state.
```useState``` is helps us to controlled state and set initial value being updated along with React live cycle.

So for this case, I'm trying to set initial value for nextSectionIndex is 0. and this state will be use for calling index of Array and it's will be updated via ```setNextSectionIndex``` function.

File: akvo/rsr/spa/app/modules/editor/project-init-handler.js (hunk 3 above)
Comment (ifirmawan) [reply akvo/rsr/spa/app/modules/editor/project-init-handler.js:76]: Got it! will do.

File: akvo/rsr/spa/app/modules/editor/project-init-handler.js (hunk 2 above)
Comment (ifirmawan) [reply akvo/rsr/spa/app/modules/editor/project-init-handler.js:73]: yes yes, I'm in. Okay @MichaelAkvo I'm working to fix all feedbacks

File: akvo/rsr/spa/app/modules/editor/project-init-handler.js (hunk 1 above)
Comment (ifirmawan) [reply akvo/rsr/spa/app/modules/editor/project-init-handler.js:82]: Because some field arrays are rendered first before the data state is ready and I can't find an area to indicate the state is ready. So, that's why I delayed to set fetched status as true when we got success response from backend.

File: akvo/rsr/spa/app/modules/editor/project-init-handler.js (hunk 4 above)
Comment (MichaelAkvo) [reply akvo/rsr/spa/app/modules/editor/project-init-handler.js:18]: I see. Thanks for explaining

File: akvo/rsr/spa/app/modules/editor/project-init-handler.js [hunk 5]
```diff
@@ -11,10 +11,14 @@ const insertRouteParams = (route, params) => {
   })
   return route
 }
-let prevParams
+
 let sectionIndexPipeline = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]
+const SECTION_RESULTS_INDICATORS = 4;
+const SECTION_COUNT = 11;
```
MichaelAkvo, line 17: It should be worth changing this to something like

```js
const SECTION_RESULTS_INDICATORS = 4;

let sectionIndexPipeline = [1, 2, 3, SECTION_RESULTS_INDICATORS, 5, 6, 7, 8, 9, 10, 11];
const SECTION_COUNT = sectionindexPipeline.length + 1;
```

If you know the name of the sections, then you could just insert them all for later use.

MichaelAkvo (changes requested):

Issue #5140, opened 2022-10-10
Stabilize period update aggregations


MichaelAkvo:
# TODO

 - @zuhdil  Move all code into a cronjob (akvo.rsr.models.result.indicator_period.IndicatorPeriod.recalculate_period)
     - general function in `akvo-rsr/akvo/rsr/usecases/period_update_aggregation.py:aggregate()`
 -  @zuhdil Wrap aggregation process in database transaction
 - @MichaelAkvo  Write command
    -  `AggregationTask` model (reuse akvo.rsr.models.iati_validation_job.IatiActivityValidationJob)
      - id
      - period_data: PeriodData
      - status: SCHEDULED, RUNNING, DONE, FAILURE, REQUIRES_INTERVENTION
      - pid: process ID of the task
      - created_at
      - started_at
      - retry_count

PR #5147, opened 2022-10-12
[#5140] stabilize period update aggregations

MichaelAkvo:
# TODO / Done

 - [ ] Add aggregation job management feature
 - [ ] Split out aggregation to a function
 - [ ] Call aggregation in a command
 - [ ] schedule aggregation job on IndicaterPeriod update

# Test plan

 - [ ] unit tests
 - [ ] Local tests
 - [ ] Deploy to test environment

Related to #5140


akvo/rsr/usecases/jobs/aggregation.py
```diff
@@ -0,0 +1,107 @@
...
+            except Exception as e:
+                scheduled_job.mark_failed()
+                email_failed_job_owner(scheduled_job, str(e))
+
+
+@atomic
+def handle_failed_jobs():
+    """Identify failed jobs, notify owners, and reschedule them"""
+    fail_dead_jobs()
+    failed_jobs = get_failed_jobs()
```
zuhdil, line 79: The failed jobs that are not caused by a dead process could also be included

akvo/rsr/models/cron_job.py
```diff
@@ -0,0 +1,52 @@
...
+        self.pid = None
+        self.save()
+
+    def mark_failed(self):
+        self.status = self.Status.FAILED
+        self.attempts = self.attempts + 1
+        self.pid = None
+        self.save()
+
+    def mark_maxxed(self):
```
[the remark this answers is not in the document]
MichaelAkvo: This is preparation for the next task where users will be able to restart jobs that have failed too often. Since we have the status already, I thought it would be good to have the accompanying method.

akvo/rsr/usecases/jobs/aggregation.py
MichaelAkvo: They are. `fail_dead_jobs()` marks the dead jobs as failed so they are included in the results of all other failed jobs to be rescheduled.

akvo/rsr/models/result/indicator_period.py
```diff
@@ -107,12 +106,6 @@ def save(self, *args, **kwargs):
             percentage = calculate_percentage(self.numerator, self.denominator)
             self.actual_value = str(percentage)
 
-        if not new_period:
-            # Check if the actual value has changed
-            orig_period = IndicatorPeriod.objects.get(pk=self.pk)
-            if orig_period.actual_value != self.actual_value:
-                actual_value_changed = True
-
```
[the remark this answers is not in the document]
zuhdil: I'm not sure I understand your question, but the main approach of the old code for aggregating values to the parent period is by detecting changes in the actual_value attribute when saving the period object. This approach for me seems complicated and hard to understand.

akvo/rsr/usecases/jobs/aggregation.py
```diff
@@ -0,0 +1,154 @@
...
+
+def execute_aggregation_jobs():
+    """
+    Call the aggregation function for each aggregation job
+    """
+    handle_failed_jobs()
+
+    scheduled_jobs = get_scheduled_jobs()
+    for scheduled_job in scheduled_jobs:
+        with atomic():
```
MichaelAkvo, line 67: I think this should be a while loop that keeps getting the first job. 

```py
while (scheduled_job := get_scheduled_jobs().first()):
```

And maybe `atomic` is unnecessary here too. Marking the job as running won't matter if the transaction ends because it will either be marked as finished or failed at the end of the transaction.

PR #5155, opened 2022-10-31
[#5140] Create command to audit project aggregation values

zuhdil:
# TODO / Done

 - [x] Command to audit project aggregation values
 - [x] Cronjob to audit MWC program aggregation

# Test plan

 - [ ] Manual testing

## Happy path

 run `./manage.py audit_project_aggregation 9062`


akvo/rsr/management/commands/audit_program_aggregation.py
```diff
@@ -0,0 +1,52 @@
...
+    if period.indicator.measure == PERCENTAGE_MEASURE:
+        contrib_numerator, contrib_denominator = aggregation.sum_contributed_percentage_value(period)
+        numerator = ensure_decimal(numerator) + ensure_decimal(contrib_numerator)
+        denominator = ensure_decimal(denominator) + ensure_decimal(contrib_denominator)
+        value = calculate_percentage(numerator, denominator)
+    else:
+        value = ensure_decimal(value) + aggregation.sum_contributed_unit_value(period)
+
+    try:
+        assert ensure_decimal(period.actual_value) == value
```
MichaelAkvo, line 21: How many places after the comma are used for comparison? [By default it's 28](https://docs.python.org/3/library/decimal.html?highlight=compare%20decimal#decimal.DefaultContext) which might be a lot and yield invalid results.

Also why use `assert` here instead of `if`?

akvo/rsr/management/commands/audit_program_aggregation.py
```diff
@@ -0,0 +1,52 @@
...
+            print("Program not found")
+            return
+
+        running_quantitative_periods = IndicatorPeriod.objects\
+            .filter(indicator__result__project=program, indicator__type=QUANTITATIVE)\
+            .exclude(Q(actual_value__isnull=True) | Q(actual_value__exact=''))
+
+        count = 0
+        for period in running_quantitative_periods:
+            count += audit_period_aggregation(period)
```
MichaelAkvo, line 51: I think it would be good to put this into a usecase module with a test. Otherwise when we merge this into `master`, it might break and we won't notice it. (I can't remember if `ProjectHierarchy` still exists as a model)

MichaelAkvo (changes requested):

akvo/rsr/management/commands/audit_program_aggregation.py
zuhdil: fixed.

akvo/rsr/management/commands/audit_program_aggregation.py
zuhdil: `ProjectHierarchy` is a model used to indicate that a project is a program. It's not a proper name but that's what we have.

I've updated the code to accept `project_id` as a parameter so it's not limited to just programs.

PR #5173, opened 2022-12-06
[#5170] Implement command to switch indicator to cumulative reporting

### LINKED ISSUE #5170: Feature Request: Script to bulk change indicators to cumulative reporting
### What are you trying to do?

I want to change indicators on a project to cumulative report indicators

### Describe the solution you'd like

Create a Django management command

### Have you consider alternatives?

_No response_

### Additional context

_No response_

zuhdil:
# TODO / Done

 - [x] Implement Django command

# Test plan

 - [x] Unit test
 
---

Closes #5170

### DIFF
--- a/akvo/rsr/management/commands/switch_indicators_to_cumulative_reporting.py
+++ b/akvo/rsr/management/commands/switch_indicators_to_cumulative_reporting.py
@@ -1,0 +1,59 @@
+import argparse
+from io import TextIOBase
+from typing import List
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from akvo.rsr.models import Project, Indicator
+from akvo.rsr.models.result.utils import QUANTITATIVE
+
+
+class Command(BaseCommand):
+    help = "Switch all quantitative indicators of the given project and its descendants to cumulative reporting"
+
+    def add_arguments(self, parser: argparse.ArgumentParser):
+        parser.add_argument("project_id", type=int)
+        parser.add_argument(
+            "--exclude", nargs="+", help="ID of indicators that should not be changed"
+        )
+        parser.add_argument(
+            "--dry-run", action="store_true", help="Don not actually apply the changes"
+        )
+
+    def handle(self, *args, **options):
+        project = Project.objects.get(id=int(options["project_id"]))
+        excludes = [int(id) for id in options["exclude"]] if options["exclude"] else []
+        runner = CommandRunner(self.stdout, project, excludes, options["dry_run"])
+
+        try:
+            runner.run()
+        except InterruptedError:
+            self.stdout.write("Changes not applied\n")
+        else:
+            self.stdout.write("Changes applied\n")
+        self.stdout.write("DONE!\n")
+
+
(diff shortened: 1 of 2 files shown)


akvo/rsr/management/commands/switch_indicators_to_cumulative_reporting.py
```diff
@@ -0,0 +1,59 @@
...
+        self.stdout = stdout
+        self.project = project
+        self.excludes = excludes
+        self.dry_run = dry_run
+
+    @transaction.atomic
+    def run(self):
+        indicators = Indicator.objects.filter(
+            result__project=self.project, type=QUANTITATIVE
+        )
```
MichaelAkvo, line 49: I just tested this and it doesn't work with descendants that defined their own quantitative indicators. They won't be picked up in the inheritance chain.

A possible fix could be

```python
indicators = Indicator.objects.filter(result__project__in=self.project.descendants, type=QUANTITATIVE)
```

To test that it works, I ran

```sh
./manage.py switch_indicators_to_cumulative_reporting 9062
./manage.py shell
```
```python
from akvo.rsr.models import *
QUANTITATIVE=1
mwc = Projects.objects.get(id=9062)
Indicator.objects.filter(result__project__in=mwc.descendants(), type=QUANTITATIVE, cumulative=False).count())
# 159

# get those without a parent
Indicator.objects.filter(result__project__in=mwc.descendants(), type=QUANTITATIVE, cumulative=False, parent_indicator__isnull=True).count()
# 59
```

They are across multiple projects.

akvo/rsr/management/commands/switch_indicators_to_cumulative_reporting.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class Command(BaseCommand):
+    help = "Switch all quantitative indicators of the given project and its descendants to cumulative reporting"
+
+    def add_arguments(self, parser: argparse.ArgumentParser):
+        parser.add_argument("project_id", type=int)
+        parser.add_argument(
+            "--exclude", nargs="+", help="ID of indicators that should not be changed"
+        )
```
MichaelAkvo, line 17: The client has responded and doesn't require this anymore. Due to my other comment (uninherited quantitative indicators), I think the easiest would be to simply remove this option. Otherwise each indicator would have to be checked to ensure that it's not in the descendants of an excluded indicator.

MichaelAkvo (changes requested):

akvo/rsr/management/commands/switch_indicators_to_cumulative_reporting.py
zuhdil: Fixed

akvo/rsr/management/commands/switch_indicators_to_cumulative_reporting.py
zuhdil: Fixed

MichaelAkvo (approved):
works fine!

PR #5195, opened 2023-01-12
[#5194] indicator periods cumulative summary value

### LINKED ISSUE #5194: Bug: The summary value of the indicator periods is not calculated cumulatively
### What were you doing?

The summary value of indicator periods with cumulative reporting enabled is not calculated cumulatively, this applies to the project's results overview page and some reports.

### What should've happened?

The summary value of indicator periods with cumulative reporting enabled should be calculated cumulatively

### My environment

_No response_

### Additional context

_No response_

zuhdil:
# TODO / Done

 - [x] Fix frontend
 - [x] Fix  reports

# Test plan

 - [ ] Manual testing
 
Fixes #5194

### DIFF
--- a/akvo/rsr/views/py_reports/kickstart_word_report.py
+++ b/akvo/rsr/views/py_reports/kickstart_word_report.py
@@ -81,14 +81,13 @@ def build_log_frame(project_view):
                     previous_indicator = indicator.title
                     current_indicator = indicator.title
                 disaggregations = []
-                for d in period.disaggregations.all():
+                for d in period.disaggregations.values():
                     disaggregations.append({
-                        'label': str(d.dimension_value),
-                        'value': d.value
+                        'label': str(d.get('type', '')),
+                        'value': d.get('value', None)
                     })
                 if len(disaggregations):
                     has_disaggregations = True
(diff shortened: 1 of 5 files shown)


akvo/rsr/views/py_reports/kickstart_word_report.py
```diff
@@ -81,14 +81,13 @@ def build_log_frame(project_view):
                     previous_indicator = indicator.title
                     current_indicator = indicator.title
                 disaggregations = []
-                for d in period.disaggregations.all():
+                for d in period.disaggregations.values():
                     disaggregations.append({
-                        'label': str(d.dimension_value),
-                        'value': d.value
+                        'label': str(d['type']),
+                        'value': d['value']
```
MichaelAkvo, line 87: I think it would be good to use defensive programming here and use `.get`

MichaelAkvo (changes requested):
A small change to counter possible events

akvo/rsr/views/py_reports/kickstart_word_report.py
zuhdil: Fixed

Issue #5211, opened 2023-02-01
Bug: Cumulative indicators have incorrect aggregated value in program overview


MichaelAkvo:
### What were you doing?

Open the [program overview for Making Water Count](https://rsr.akvo.org/my-rsr/programs/9062)

![image](https://user-images.githubusercontent.com/91939214/216002896-d546bf4c-0656-46b4-86c3-a3c5558cdaa6.png)

The indicator is cumulative but the aggregated actual value over all periods is the sum of the periods.

### What should've happened?

The aggregated value over all periods for cumulative indicators should be the value of the last period with a value.

PR #5224, opened 2023-02-14
[#5211] Fix incorrect aggregated value pg overview

ifirmawan:
# TODO / Done

 - [x] 1. Fix incorrect aggregated value in the program overview by getting an actual value from the current period when the cumulative status is true.
 - [x] 2. Fix handling of rescheduling endpoints to use a list of aggregation jobs.

# Test plan

 - [x] Revalidate aggregated value in indicator level (cumulative and non-cumulative).
 - [x] Re-run the job when any aggregation jobs were failed/max attempts.
 - [x] Update some value in any projects and run aggregations jobs.

### DIFF
--- a/akvo/rest/views/project_overview.py
+++ b/akvo/rest/views/project_overview.py
@@ -373,6 +373,7 @@ def project_result_overview(request, project_pk, result_pk):
                     'unit' if i.measure == '1' else 'percentage' if i.measure == '2' else None),
                 'periods': _drilldown_indicator_periods_contributions(i, aggregate_targets),
                 'disaggregation_targets': _transform_disaggregation_targets(i),
+                'cumulative': i.cumulative,
             }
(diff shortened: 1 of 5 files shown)


ifirmawan:
Hi @MichaelAkvo, FYI for the `target value` we reviewed yesterday. I found value coming from the backend side
https://github.com/akvo/akvo-rsr/blob/master/akvo/rest/views/project_overview.py#L370
and the front-end show as it is
https://github.com/akvo/akvo-rsr/blob/98bee7d0ea5d23033ed8d3d676d735b0a374b73a/akvo/rsr/spa/app/modules/program/indicator.jsx#L15

MichaelAkvo:
Everything works as expected :)

MichaelAkvo (approved):

MichaelAkvo:
Related to #5211

PR #5242, opened 2023-03-06
[#5241] Fix incorrect cumulative periods aggregation values

### LINKED ISSUE #5241: Bug: Cumulative indicator aggregation is incorrect when the previous period gets updated
### What were you doing?

Given there are three periods of a cumulative indicator, the first and second periods have updates, and the total value of the second period is carried over to the third period. 

When the first period gets an additional update, the total value of the first period is carried over to the second and third periods, and the second period is ignored.

### What should've happened?

The additional update should not change the total value of the second period and not be carried over to the third period.

### My environment

_No response_

### Additional context

_No response_

zuhdil:
# TODO / Done

 - [x] Fix incorrect cumulative periods aggregation values when the previous period gets updated

# Test plan

 - [x] Unit test
 - [x] Manual test
 
---

Closes #5241

### DIFF
--- a/akvo/rsr/models/result/utils.py
+++ b/akvo/rsr/models/result/utils.py
@@ -7,7 +7,7 @@
 from decimal import Decimal
 
 from django.apps import apps
-from django.db.models import Max, Q
+from django.db.models import Q
 from akvo.utils import rsr_image_path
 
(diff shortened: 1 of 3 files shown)


MichaelAkvo (approved):
Looks good to me!

Issue #5263, opened 2023-03-22
Bug: django q probe times out on kubernetes


MichaelAkvo:
### What were you doing?

Starting up the application on kubernetes. 

![image](https://user-images.githubusercontent.com/91939214/226901666-6d9aa1b1-8006-403c-95ef-d35727864e20.png)

### What should've happened?

The readiness probe shouldn't time out so frequently.

PR #5264, opened 2023-03-22
[#5263] Provide localhost server for django-q probe

MichaelAkvo:
# TODO / Done

 - [x] Add localhost server for django-q probe
 - [x] Add django-q probe using `curl`
 - [x] Use probe in docker-compose
 - [x] Use probe in k8s

# Test plan

 - [x] Start up docker-compose

Related to #5263


scripts/probes/worker.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# The status is provided by a local server started by
+#  ./manage.py django_q_probettp
+
+status=$(curl -s localhost:8080)
+echo "returned status ${status}"
+[[ $status == "Started" ]] || [[ $status == "Idle" ]]
```
MichaelAkvo, line 7: @zuhdil  could you please add the "Working" status here?

https://github.com/Koed00/django-q/blob/15155c7a99ea85a092b1d11da730ba0c4fa98ffc/django_q/conf.py#L197

PR #5320, opened 2024-01-04
Make script to get list of users related to and organisation

zuhdil:
---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1206278094409187


zuhdil:
This is already implemented in https://github.com/akvo/akvo-rsr/pull/5322/commits/22bcd379886ddfaf3a2606b608bb68d73db5c196

